Issue #14, opened 2026-04-03
Roll Dice UI


SoupCube:
### User Story
As a player, I want to roll a dice, that determines the outcome of my turn.

### Acceptance Criteria
- Active player can trigger a dice roll at the start of their turn
- The result is displayed clearly
- All players see the same dice result
- A short animation or visual feedback is shown when rolling

### Notes
- Backend should generate the random number (single source of truth)
- Requires real-time sync

#### Related Issues
- https://github.com/SE2-Machi-Koro/Server/issues/43

Issue #12, opened 2026-04-03
Joining Lobby UI


SoupCube:
### User Story 
As a player, I want to join a game lobby using a code so that I can connect to a specific game session

### Acceptance Criteria
- A text input field is available for entering a lobby code
- A “Join Lobby” button is available
- Player can enter a code
- Valid code → player joins the corresponding lobby
- Invalid code → error message is shown
- While joining, a loading state is displayed (optional)
- After successful join, player is navigated to the lobby screen

### Notes
- Requires backend validation of lobby codes
- Consider input validation (length, allowed characters)

#### Related Issues
- https://github.com/SE2-Machi-Koro/Server/issues/38
- https://github.com/SE2-Machi-Koro/Server/issues/40

PR #79, opened 2026-04-25
feat: add navigation from start screen to game screen with dice rolling

### LINKED ISSUE #46: Implement Dice Roll Button Story
As an active player, I want to see a "Roll Dice" button that I can click at the start of my turn.
Acceptance Criteria:

- Button is only visible/clickable for the active player
- Button is disabled when it is not the player's turn
- Click sends a roll event to the server via WebSocket

kutschera13:
- Add "Spiel starten" button on StartScreen (visible when connected)
- Navigate to GameScreen after game start
- GameScreen shows "Würfeln" button that sends roll event via WebSocket

Out of scope (separate ticket):
- Active player detection (who's turn is it)
- Button disabled when not player's turn


FabianSolutions (dismissed):
- GameScreenStateTest.kt is broken. Test file was overwritten with a duplicate data class GameScreenState in the same package as the production class. No @Test methods remain. Likely won't compile; either delete the file or restore it.                                                     
- Lifecycle bug. GameViewModel.onCleared() calls wsClient.disconnect(), but the same webSocketClient is shared with StartScreenViewModel and MainActivity.onStop already handles disconnect. Navigating back will kill the connection. Remove the VM-level disconnect.
- Package vs. directory mismatch in 3 new files:
  - ui/game/GameScreen.kt declares package …ui
  - ui/game/GameViewModel.kt declares package …viewmodel
  - domain/model/state/DiceRollResult.kt declares package …network.websocket

Should fix

- 2-dice rendering is broken. Parser builds DiceRollResult(dice = listOf(total), total = total). With total > 6 the UI renders "?". Parse each die from the payload.
- canRollTwo is dead. unlockTwoDice() is never called — the 1/2 dice picker never shows. Wire it or remove it.
- Hardcoded playerId = "android-client". PR body acknowledges this is out of scope; at minimum add a TODO.
- No tests for GameViewModel, GameScreen, or the new OkHttpWebSocketClient methods. The interface gained two members — worth grepping for other WebSocketClient fakes that may break.
- Refactor noise. Several unrelated one-line collapses in OkHttpWebSocketClient belong in a separate cleanup PR.

Nits

- ~9 files lost their trailing newline.
- Aligned-with-spaces var declarations in GameScreen.
- Hardcoded German strings (out of scope project-wide?).
- catch (e: Exception) in handleDiceResult is too broad.
- GameScreenState.gamePhase default is ROLL_DICE while StartScreenState.gamePhase default is NONE.

Solid

- Nav flow and the gamePhase != NONE switch are clean.
- WebSocketContract extended properly — no magic strings elsewhere.
- Subscription placed at STOMP CONNECT, alongside the public topic.

request changes — the broken test file and the disconnect bug are the real problems; the rest is cleanup.

eles17 (changes requested):
**Please don't merge this yet.** I would like to first resolve the parent issue and sub-issues with Lea, so the PR is assigned to the correct sub-issue.

Right now, if this is merged, the whole parent issue would be marked as done, although this PR only introduces functionality for part of a sub-issue.

Moritz2k1:
I would like to merge this PR before the others, so can you please resolve the issues, implement tests for code coverage and get the code working. Me and Fabian are also willing to help to speed up the process

valentinaschiavon99:
Please @kutschera13 when u have time solve these conflicts

eles17 (changes requested):
Thanks for pushing through all the conflict resolution here. I can see the branch went through a lot of cleanup and syncing with main, so I think the current state may have drifted from the original PR scope.

I would not merge this yet, but the core issue is pretty focused: the current diff adds the WebSocket/ViewModel plumbing for dice rolling, but the user-facing button story from #46 is not wired anymore. `GameScreenViewModel.rollDice()` exists, but `GameScreen` does not receive a callback, does not show a “Würfeln” button, and does not display `diceResult`.

It looks like the dice button existed in an earlier version of this branch, but it was implemented against an older GameScreen/GameViewModel structure. After main introduced the current state-driven GameScreen, the old dice UI was not migrated into the new GameScreen(state = ...) API. During conflict resolution/refactoring, the old GameViewModel was removed and the final PR kept only the WebSocket/ViewModel plumbing.

So I think we have two good options:

1. Keep this PR small and rename/rescope it as dice-roll WebSocket plumbing only, then create/follow up with a separate PR for the actual button UI.
2. Finish #46 in this PR by wiring the button/result display into `GameScreen` and adding a couple of tests.

I also recommend adding tests for:
- `OkHttpWebSocketClient.rollDice()` sending the correct STOMP frame
- incoming `ROLL_DICE` messages updating `diceResult`
- `GameScreenViewModel.rollDice()` forwarding the dice count
- `GameScreenViewModel` collecting `diceResult`

The build is green on my side (`test` and `assembleDebug` pass), so this is not a broken-branch situation. It is mainly a scope/alignment issue between #46, parent #14, and the current final diff.

Let us know what option you want to do and reach out if you need any help.

eles17 (dismissed):
I think this PR is mergeable for the current dice-roll scope.

It wires the roll button through `GameScreen` -> `GameScreenViewModel` -> `WebSocketClient`, sends the `ROLL_DICE` frame, receives `ROLL_DICE` results, and displays the dice result in the game screen. Unit tests and Gradle build are green.

The existing `showLobbyScreen` / lobby navigation behavior and related `AppRootTest` problem should be handled separately in the navigation/lobby issue #12 as separate sub-issue @Ivona-Goranova , not as a blocker for this dice-roll PR.

eles17 (approved):
Approved from my side for the current dice-roll scope.

Issue #106, opened 2026-05-01
Gate lobby/game UI on authenticated state


FabianSolutions:
### Goal

Today the start screen shows lobby/game actions regardless of whether the user is logged in. Once Server #160 lands, game-mutating WebSocket endpoints will reject unauthenticated traffic — meaning an unauthenticated user who reaches the game screen will just see error frames. This issue gates the lobby/game UI on the authenticated state and handles auth rejection at the network layer.

### Tasks

- [ ] On `StartScreen`, hide or disable lobby/game actions when `state.loggedInAs == null`. The Register + Login buttons stay; non-auth actions like "Rules" can stay since they're harmless. Lobby buttons (added by Client #82, future ones) must be gated.
- [ ] When an unauthenticated user attempts a game action (if any reachable), surface a clear "Please log in" prompt — open `LoginDialog` directly or show a snackbar.
- [ ] Handle WebSocket auth-rejection at the network layer: if the server returns a STOMP `ERROR` frame indicating auth failure (e.g. expired/invalid token on a previously-authenticated connection), force `SessionManager.signOut()` and surface "Your session has expired, please log in again" via a snackbar or dialog.
- [ ] Compose UI test: unauthenticated start screen hides game actions; authenticated start screen shows them.
- [ ] Test the auth-rejection path: simulate a STOMP ERROR frame with an auth-expired body via the existing `FakeWebSocketFactory`, assert `SessionManager.signOut()` is called and the UI transitions back to unauthenticated.

### Out of scope

- Refresh tokens / silent re-auth.
- Per-action authorization (e.g. only host can start) — server's concern.
- Anonymous-game support — out of scope for the SE2 auth flow.

### Acceptance

- Unauthenticated start screen shows Register + Login + minimal context (Rules), no game actions.
- Authenticated start screen shows the lobby flow + Logout.
- A WebSocket auth-rejection (token cleared server-side, expired, etc.) cleanly transitions the UI back to unauthenticated and clears `SessionManager`.

### Dependencies

- Blocks on Server #160 merging (so the auth-rejection error frame is real and we can test against the actual contract).
- The static UI gating part can land before #160; the rejection-handling part needs the server contract.

### Related

Counterpart server: SE2-Machi-Koro/Server#160
Source: Client PR #92's "Out of scope" section called this out.
Parent: SE2-Machi-Koro/Server#157

Issue #105, opened 2026-05-01
Persist session token across app cold-starts


FabianSolutions:
### Goal

`SessionManager` (Client #90 / PR #92) holds the session token in memory only. The token survives screen rotation but not an app cold-start — every app launch forces the user to log in again. This issue persists the token so a logged-in user stays logged in across launches.

### Tasks

- [ ] Pick the storage layer:
  - **Preferences DataStore** (recommended): modern, async by default, plain text on the device. Acceptable for SE2 scope where the threat model doesn't include rooted-device adversaries.
  - **EncryptedSharedPreferences**: encrypted at rest, slightly heavier. Probably overkill for SE2; flag as the upgrade path if the threat model shifts.
- [ ] On app launch, read the persisted token (if any) and call `SessionManager.signIn(...)` with it. The existing `StartScreenViewModel` collector then picks it up and the UI shows the authenticated state.
- [ ] On `SessionManager.signIn(...)`, write to persistent storage.
- [ ] On `SessionManager.signOut()`, clear the persistent storage entry.
- [ ] Decide validation strategy: don't add a `GET /auth/me` server endpoint just for this — instead, rely on the next authenticated WebSocket call to fail with the auth-rejection frame, which the auth-gate issue will handle by surfacing "session expired" and signing out.
- [ ] Unit tests on the persistence wrapper (read / write / clear).
- [ ] Integration test on the cold-start path (persisted token → user is logged in after launch).

### Out of scope

- Token expiration / refresh — server doesn't expire tokens yet (Server #158 deferred this).
- Multi-account / "switch user".
- Encrypted storage beyond plain DataStore.

### Acceptance

- After a successful login, killing and relaunching the app preserves the logged-in state.
- After logout, the persisted token is cleared and a relaunch shows the unauthenticated UI.
- A persisted-but-now-invalid token (e.g. server cleared it via DB reset) gracefully transitions to logged-out on the next authenticated call rather than crashing.

### Dependencies

- None blocking. Could land before or after WebSocket auth (Client issue for #159 counterpart).

### Related

Source: Client PR #92's "Out of scope" section called this out.
Parent: SE2-Machi-Koro/Server#157

PR #109, opened 2026-05-01
feat: #104 send session token in WebSocket STOMP CONNECT headers

### LINKED ISSUE #104: Send session token in WebSocket STOMP CONNECT headers when authenticated
### Goal

Wire the session token into the WebSocket layer so the Android client can establish authenticated STOMP sessions. Server #159 (PR #179) ships a strict-reject interceptor on the inbound STOMP channel that requires `Authorization: Bearer <token>` on every CONNECT and rejects anonymous handshakes outright. Without a corresponding client change, every WebSocket connection from the Android app fails at CONNECT.

### Tasks

- [ ] `OkHttpWebSocketClient` takes a `SessionStateHolder` constructor argument (or reads `SessionManager.session` directly) and adds `Authorization: Bearer <token>` as a STOMP native header on the CONNECT frame. Header name and prefix must match Server #159's `StompAuthChannelInterceptor` exactly (`AUTH_HEADER = "Authorization"`, `BEARER_PREFIX = "Bearer "`).
- [ ] **Defer the WebSocket connect until the session is non-null** — do NOT attempt anonymous connections. The strict-reject server would close them at CONNECT and the user would see a misleading "connection error" before they've even logged in.
- [ ] Read the token at `onOpen` time (the moment the STOMP handshake begins), not at `connect()` time. This closes the race where the user logs out between `connect()` and the WS handshake completing — without this, a stale token would be sent and rejected.
- [ ] If the session vanishes between `connect()` and `onOpen`, close the underlying socket cleanly without sending CONNECT.
- [ ] When the session changes (login → logout, or vice versa), tear down the current WebSocket and reconnect with the new state. Use a coroutine collector on `SessionManager.session` (e.g. a `LaunchedEffect` in `MainActivity`).
- [ ] `disconnect()` must be truly idempotent — a no-op when there was no socket. Otherwise the session-driven collector's initial emission of `null` (cold start with no session) flips status from `IDLE` to `DISCONNECTED` and the start screen shows a misleading "Connection status: disconnected" before the user has tried anything.
- [ ] Unit tests:
  - CONNECT frame includes the `Authorization: Bearer …` header when a session is present.
  - `connect()` is a no-op (no socket opened, status stays `IDLE`) when there's no session.
  - The token used in CONNECT is the one current at `onOpen`, not at `connect()` (locks down the read-at-handshake-time contract).
  - The socket closes cleanly without sending CONNECT if the session vanishes between `connect()` and `onOpen`.
  - `disconnect()` is a no-op when never connected.

### Out of scope

- **Token persistence across cold-starts** (DataStore / EncryptedSharedPreferences) — Client #105.
- **Server-side auth-rejection UX** (force-logout on rejection mid-session) — Client #106.
- **Mid-game token expiration** — Server #158 doesn't expire tokens yet.
- **Narrow login/logout/login race during the WS-close window** — would need a closing-generation counter to handle properly. Defer until observed in practice.

### Acceptance

- Logged-in client opens the WS with `Authorization: Bearer <token>` set; server log confirms `WS CONNECT accepted for user '<x>'`.
- Logged-out client does NOT open a WS at all; status stays `IDLE`. No "connection error".
- Login while foregrounded triggers a connect with the new token.
- Logout while foregrounded triggers a clean disconnect.
- Re-login (logout → login) reconnects with the fresh token.

### Dependencies

- Blocked by Server PR #179 (issue #159) merging — the server must require the header before this lands so the auth flow is actually enforced.
- Implementation can start in parallel against Server #179's documented contract.

### Related

Counterpart server: SE2-Machi-Koro/Server#159, SE2-Machi-Koro/Server#179
Parent flow: SE2-Machi-Koro/Server#157
Follows-on: persistent token storage (Client #105) and auth-gating UI (Client #106).

---

**Note:** Earlier versions of this issue suggested *"skip the auth header when the session is null"* and *"server should permit anonymous"*. That was based on a softer server design. Server #159 / PR #179 chose strict-reject instead — every connecting client must already be authenticated. The tasks above reflect the current contract.

FabianSolutions:
Closes #104. Pairs with Server PR SE2-Machi-Koro/Server#179.

⚠️ **Do not merge before Server #179 is on `main`** — the new client will send `Authorization: Bearer …` on every STOMP CONNECT, but the contract that requires it (and rejects connections that don't) is on the server side. If we merge this first, nothing breaks (the header just gets ignored), but the auth flow remains insecure. The natural deploy order is: Server #179 lands → Client #104 lands.

## Summary
- `OkHttpWebSocketClient` now takes a `SessionStateHolder` constructor argument and is auth-aware.
- The CONNECT frame includes `Authorization: Bearer <token>` as a STOMP native header — exactly matching Server #179's `accessor.getFirstNativeHeader("Authorization")` + `BEARER_PREFIX = "Bearer "` contract.
- The token is read inside `onOpen`, not at `connect()` time, so a logout during the WS handshake doesn't ship a stale token.
- `connect()` short-circuits when there's no session — no socket opened, no status transition. The strict-reject server would reject anonymous CONNECT anyway; this avoids the misleading "connection error" the user would otherwise see when they simply aren't logged in yet.
- `disconnect()` is now truly idempotent (no-op when there was no socket). Without this, the new `LaunchedEffect` (commit 2) would flip status from `IDLE` to `DISCONNECTED` on every cold-start with no session.
- `MainActivity` reacts to login/logout while foregrounded via a `LaunchedEffect` collecting `SessionManager.session` and calling `connect()` / `disconnect()` accordingly. Activity-lifecycle behaviour (`onStart` / `onStop`) is unchanged.

## Header contract
`Authorization: Bearer <session-token>` as a STOMP native header on the CONNECT frame. Spelled out as constants on both sides:
- Client: `OkHttpWebSocketClient.AUTH_HEADER` + `BEARER_PREFIX`.
- Server: `StompAuthChannelInterceptor.AUTH_HEADER` + `BEARER_PREFIX`.

## What's intentional
- **Strict-reject parity with Server #179.** No anonymous WS connections — if there's no session, we don't even open the socket. Aligns the client UX with the server's behaviour: no fake "connection error" before login, no doomed handshake attempt.
- **Token read at `onOpen`, not at `connect()`.** Tested explicitly (`connectFrameUsesCurrentSessionTokenAtHandshakeTime`).
- **Vanished-session-during-handshake closes cleanly.** Tested (`handshakeClosesCleanlyIfSessionVanishedBetweenConnectAndOnOpen`).
- **No changes to the `WebSocketClient` interface.** Auth concern stays impl-internal; ViewModels and tests that mock the interface are unaffected.

## Out of scope
- **Token persistence across cold-starts** — Client #105.
- **Auth-rejection UX** (force-logout when server rejects mid-session) — Client #106.
- **Refresh tokens / token rotation** — server doesn't expire tokens yet.
- **Narrow login/logout/login race during the WS-close window** — described in self-review; the obvious fix (clear socket synchronously after vanished-session close) creates a worse problem (late `onClosed` clobbering a fresh CONNECT's status). Properly handling this needs a closing-generation counter that ignores callbacks from abandoned sockets. File a follow-up if ever observed in production.

## Test plan
- [x] `./gradlew :app:compileDebugKotlin :app:compileDebugUnitTestKotlin --warning-mode all` — clean, no warnings, no deprecation notes.
- [x] `./gradlew :app:testDebugUnitTest` — all unit tests pass: 4 new auth-header tests, 1 new idempotent-disconnect test, all existing tests green via the `newClient(...)` helper.
- [x] `./gradlew :app:jacocoTestCoverageVerification` — 80% coverage gate satisfied.
- [x] `./gradlew :app:lintDebug` — Android Lint passes.
- [x] **Manual end-to-end** against Server PR #179's branch:
  - Run server on Server branch `159-websocket-auth-principal` (or `main` once #179 merges).
  - Run client on this branch from Android Studio.
  - **Cold launch, no login** → no WS connect, "Connection status: idle" stays.
  - **Login** → WS connects, server log: `WS CONNECT accepted for user '<x>'`. Status flips to `CONNECTED`.
  - **Logout** → WS disconnects, status flips to `DISCONNECTED`.
  - **Re-login** → WS reconnects with the new token (verify in pgAdmin: `users.session_token` is the new UUID).
  - **Background → foreground** while logged in: status briefly drops to `DISCONNECTED` then back to `CONNECTED`.

## Heads-up for reviewers
- The old test `disconnectWithoutActiveSocketStillUpdatesStatus` was deleted — it asserted the *opposite* of the new idempotent contract. Replaced by `disconnectWhenNeverConnectedIsNoOpAndDoesNotTransitionStatus` with an explanatory comment.
- The Client #104 issue body still says *"skip the auth header when the session is null"* — that's stale per Server #179's strict-reject design. I'll update the issue body after this PR opens (deferred so the URL in the description above stays valid).

## Commit split (2)
1. `feat: #104 add SessionStateHolder to OkHttpWebSocketClient and Authorization header` — pure client-layer change. Constructor + auth header + token-at-onOpen + idempotent disconnect + 5 new tests + `MainActivity` constructor wiring. Compiles + tests pass independently.
2. `feat: #104 drive WebSocket lifecycle from SessionManager session changes` — `MainActivity` `LaunchedEffect`. Adds the foreground reactive transitions on top of the existing activity-lifecycle path.

### DIFF
--- a/app/src/main/java/com/machikoro/client/MainActivity.kt
+++ b/app/src/main/java/com/machikoro/client/MainActivity.kt
@@ -25,7 +25,10 @@ import com.machikoro.client.ui.theme.ClientTheme
 
 class MainActivity : ComponentActivity() {
     private val webSocketClient by lazy {
-        OkHttpWebSocketClient(websocketUrl = AppConfig.websocketUrl)
+        OkHttpWebSocketClient(
+            websocketUrl = AppConfig.websocketUrl,
+            sessionStateHolder = SessionManager,
+        )
     }
     private val authApi by lazy {
         AuthApiFactory.create(AppConfig.backendBaseUrl)
--- a/app/src/main/java/com/machikoro/client/MainActivity.kt
+++ b/app/src/main/java/com/machikoro/client/MainActivity.kt
@@ -8,6 +8,7 @@ import androidx.activity.viewModels
 import androidx.compose.foundation.layout.fillMaxSize
 import androidx.compose.foundation.layout.padding
 import androidx.compose.material3.Scaffold
+import androidx.compose.runtime.LaunchedEffect
 import androidx.compose.runtime.collectAsState
 import androidx.compose.runtime.getValue
 import androidx.compose.ui.Modifier
@@ -58,6 +59,21 @@ class MainActivity : ComponentActivity() {
             val registerDialogState by registerDialogViewModel.state.collectAsState()
             val loginDialogState by loginDialogViewModel.state.collectAsState()
             val logoutState by logoutViewModel.state.collectAsState()
+
+            // Drive WebSocket lifecycle from session changes during the foreground.
+            // onStart/onStop handle the activity-lifecycle case; this handles the
+            // user-logs-in-or-out-while-app-is-open case. connect() and disconnect()
+            // are both idempotent so it's safe to call them on every emission.
+            LaunchedEffect(Unit) {
+                SessionManager.session.collect { session ->
+                    if (session != null) {
+                        webSocketClient.connect()
+                    } else {
+                        webSocketClient.disconnect()
+                    }
+                }
+            }
+
             ClientTheme {
                 Scaffold(modifier = Modifier.fillMaxSize()) { innerPadding ->
                     AppRoot(
--- a/app/src/main/java/com/machikoro/client/network/websocket/OkHttpWebSocketClient.kt
+++ b/app/src/main/java/com/machikoro/client/network/websocket/OkHttpWebSocketClient.kt
@@ -4,6 +4,7 @@ import android.util.Log
 import com.machikoro.client.domain.enums.GamePhase
 import com.machikoro.client.domain.model.state.ConnectionStatus
 import com.machikoro.client.domain.model.state.PlayerCoinState
+import com.machikoro.client.domain.session.SessionStateHolder
 import java.net.URI
 import kotlinx.coroutines.flow.MutableStateFlow
 import kotlinx.coroutines.flow.StateFlow
@@ -17,7 +18,8 @@ import org.json.JSONObject
 
 class OkHttpWebSocketClient(
     private val websocketUrl: String,
-    private val webSocketFactory: WebSocketFactory = OkHttpWebSocketFactory()
+    private val sessionStateHolder: SessionStateHolder,
+    private val webSocketFactory: WebSocketFactory = OkHttpWebSocketFactory(),
 ) : WebSocketClient {
     override val connectionStatus: StateFlow<ConnectionStatus>
         get() = mutableConnectionStatus.asStateFlow()
@@ -39,6 +41,15 @@ class OkHttpWebSocketClient(
     override fun connect() {
         synchronized(this) {
             if (webSocket != null) {
+                return
+            }
+            // Strict-reject server (#159): no session means we have nothing to
+            // present at STOMP CONNECT. Don't open a socket only to be kicked.
+            // Status stays at whatever it was — IDLE on first launch, DISCONNECTED
+            // after a logout — so the user doesn't see a misleading "connection
+            // error" when they simply aren't logged in yet.
+            if (sessionStateHolder.session.value == null) {
+                Log.d(TAG, "Skipping WS connect — no session token")
                 return
             }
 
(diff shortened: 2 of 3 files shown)


Moritz2k1 (approved):
The Server PR has been approved and merged. This implements the frontend of the session management. Looks fine to me

PR #125, opened 2026-05-04
122 extract lobbyscreen

Ivona-Goranova:
### Changes
- Extracted lobby logic from StartScreen into:
  - `LobbyScreen`
  - `LobbyScreenViewModel`
  - `LobbyScreenState`
- Cleaned up `StartScreen` to handle only authentication (login/register)
- Updated `AppRoot` and `MainActivity` to integrate the new lobby flow
- Split and updated tests:
  - Removed lobby-related tests from `StartScreenViewModelTest`
  - Added `LobbyScreenViewModelTest`

### Motivation
Improve separation of concerns and make the UI structure clearer:
- StartScreen → authentication only
- HomeScreen → create/join lobby
- LobbyScreen → player list, ready state, start game

### Notes
- I intentionally **commented out old lobby-related code instead of deleting it**, to keep context and allow easier rollback if needed.
- Navigation is still temporary (based on `lobbyCode`)
- Ready/Leave actions are not yet connected to the backend


FabianSolutions (changes requested):
Hey Ivona :) a few things to address before we can merge:
 - StartScreenViewModel.kt looks structurally off. You commented out onStartGame with /* ... */ but the Factory class ended up indented one level deeper and the final closing brace of the class got removed. Please double-check
  it still compiles, and either keep the original structure or just delete onStartGame and the old players collector outright.                                                                                                      
  - isHost is never set in LobbyScreenViewModel — it stays false forever. That means the "Spiel starten" button will always be disabled and onStartGame always returns early. The "first player = host" trick in the UI only
  controls the host icon, not the actual gating.
  - The leave button does nothing. AppRoot shows the lobby whenever lobbyCode != null, but nothing in LobbyScreenViewModel can clear lobbyCode (it lives on HomeViewModel). Right now users get stuck in the lobby. onLeaveLobby
  needs to actually clear that state.


  - Please delete the commented-out code in StartScreen.kt and StartScreenViewModel.kt (and the commented tests) instead of leaving it for "rollback". Git history is there for that, and the dead blocks make the diff really hard
  to follow.
  - The Type.kt change to bodyMedium (Normal → Bold, plus letterSpacing) affects every screen in the app, not just the lobby. If you want bold text in the lobby, override it locally.
  - Merge isReady into LobbyScreenState rather than exposing it as a second StateFlow — keeps the screen state in one place and avoids desync.
  - The placeholder actions (sendReadyToggle, sendLeaveLobby) are fine for now, but call them out explicitly in the PR description so they don't get forgotten after merge.

  Minor

  - German UI strings should move to strings.xml.
  - Add trailing newlines to the new files (LobbyScreenState.kt, LobbyScreen.kt, LobbyScreenViewModel.kt, both new tests).
  - maxPlayers = 4 is hardcoded in two places — should come from the server eventually.
  - The wildcard imports in LobbyScreen.kt don't match the rest of the codebase.
  - Other players always show as "bereit" regardless of their actual state — worth a TODO so it doesn't get lost.

  Tests

  The test coverage on the new viewmodel is solid. Once isHost is properly wired up, please uncomment and finish the onStartGame test you left as a placeholder, and add one for onLeaveLobby.

  Overall the direction is good — once the host gating, leave button, and the StartScreenViewModel structure are sorted, this is ready to go in.

Ivona-Goranova:
Thank you, Fabian, for your review. I hope I addressed everything correctly.

The `Type.kt` change was an experiment and I know that it affects every screen where `MaterialTheme.typography.bodyMedium` is used. Since we’ll likely rework the whole theme package soon as part of the upcoming UI redesign/design-system update, I left it as is for now.

FabianSolutions (approved):
Looks good — all the major points are addressed (host gating, leave flow clears lobbyCode, dead code gone, isReady merged into state, host gating tests added). Type.kt is fine to keep given the upcoming theme rework.

Two small nits you can pick up here or in a follow-up, won't block:
- German UI strings ("Spielerliste", "Spiel starten", "bereit" etc.) still inline in LobbyScreen.kt — should move to strings.xml.
- LobbyScreenState.kt, LobbyScreenViewModel.kt, and LobbyScreenStateTest.kt are missing trailing newlines.

Nice refactor, thanks!

PR #128, opened 2026-05-06
Persist session token across app cold-starts (#105)

### LINKED ISSUE #105: Persist session token across app cold-starts

FabianSolutions:
Closes #105.

## Summary
- Introduces a `SessionStorage` interface plus a Preferences DataStore-backed implementation (`DataStoreSessionStorage`).
- `SessionManager` gains `attach(storage)` and `hydrate()`; `signIn`/`signOut` write through to storage when attached. Storage failures are caught and logged so they don't escape the persistence scope.
- `MainActivity.onCreate` attaches a `DataStoreSessionStorage(applicationContext)` and launches `hydrate()` on the lifecycle scope before `setContent`. The existing `StartScreenViewModel` collector picks up the hydrated session automatically — no UI changes needed.

Split across three commits for review:
1. Add the `SessionStorage` abstraction (no callers yet)
2. Wire `SessionManager` to optional persistent storage
3. Hydrate persisted session on app start

## Acceptance map (from #105)
- ✅ Login → kill → relaunch preserves logged-in state
- ✅ Logout → relaunch shows unauthenticated UI
- ✅ Stale persisted token: hydrate emits, server rejects on next authenticated call. We do not crash. The user-visible `session expired → signOut` transition is explicitly delegated to #106 in the issue text.

## Out of scope (per issue)
- Token expiration / refresh.
- EncryptedSharedPreferences — DataStore plaintext is acceptable for the SE2 threat model; flag as the upgrade path if it shifts.
- Server `GET /auth/me` validation — invalid tokens fail naturally on the next authenticated WebSocket call.

## Tests
- New `DataStoreSessionStorageTest` — write/read/clear roundtrip + overwrite semantics, using `PreferenceDataStoreFactory.create` against a temp file.
- Extended `SessionManagerTest` — new cases for write-through on `signIn`, clear-through on `signOut`, hydrate happy path, hydrate no-op when already signed-in, hydrate with no persisted session.

Cold-start instrumentation test deferred — the unit-level `hydrate` path covers the same logic and a Compose instrumentation test on top adds little for the cost.

## Test plan
- [x] `./gradlew :app:testDebugUnitTest` — all green
- [x] `./gradlew :app:assembleDebug` — APK builds
- [x] `./gradlew :app:lintDebug` — 0 errors (only pre-existing warnings)
- [x] `./gradlew :app:jacocoTestCoverageVerification` — 80% threshold maintained
- [ ] Manual: register + login → force-stop app → relaunch → still logged in
- [ ] Manual: logout → relaunch → unauthenticated UI


Moritz2k1 (approved):
Implementation of DataStoreSessionStorage to persist the session token after new starts. Looks good

PR #129, opened 2026-05-06
Restore Register and Login buttons on StartScreen

FabianSolutions:
## Summary
The Register and Login buttons disappeared from `StartScreen` after #125 merged. An unauthenticated user only sees the title and the Rules button — the dialog state vars (`showRegisterDialog`, `showLoginDialog`) are still on the screen but nothing flips them to `true` anymore, so the dialogs are unreachable.

This restores both buttons as a centered Column using the existing `SecondaryActionButton` style, wired to the dialog state that's already in place.

## Root cause
PR #125's `cleanup: remove commented-out code` commit deleted the button blocks (which had been commented out earlier in the lobby refactor) but left the orphaned dialog scaffolding behind.

## Also fixed in this PR
`androidTest` compilation has been broken on `main` since #125 merged. Two files needed updates:

- **`AppRootTest`** — pass the new required parameters that #125 added to `AppRoot` (`lobbyScreenState`, `lobbyCode`, `loggedInAs`, `onCreateLobbyClick`, `onReadyToggle`, `onStartGame`, `onLeaveLobby`). All three test cases were missing them.
- **`StartScreenAuthStateTest`** — drop the stale cases that referenced removed `StartScreenState.isHost`/`.playerList` and the removed `onStartGame` parameter on `StartScreen`. That host-gating moved to `LobbyScreen` with #125 and is covered there. The authenticated-state cases at the `StartScreen` level are also obsolete now that auth-based routing is `AppRoot`'s responsibility.

The kept test (`unauthenticatedShowsRegisterAndLoginAndHidesLoggedInBanner`) now compiles and asserts against the restored buttons — closing the loop on what would have caught this regression.

## Test plan
- [x] `./gradlew :app:assembleDebug` — APK builds
- [x] `./gradlew :app:testDebugUnitTest` — green
- [x] `./gradlew :app:lintDebug` — 0 errors
- [x] `./gradlew :app:compileDebugAndroidTestKotlin` — green (was broken on `main`)
- [ ] Manual: launch on emulator, confirm Register and Login buttons are visible and tapping them opens the respective dialogs

### DIFF
--- a/app/src/androidTest/java/com/machikoro/client/ui/AppRootTest.kt
+++ b/app/src/androidTest/java/com/machikoro/client/ui/AppRootTest.kt
@@ -8,6 +8,7 @@ import androidx.compose.ui.test.junit4.createComposeRule
 import androidx.compose.ui.test.onNodeWithText
 import com.machikoro.client.domain.enums.GamePhase
 import com.machikoro.client.domain.model.state.GameScreenState
+import com.machikoro.client.domain.model.state.LobbyScreenState
 import com.machikoro.client.domain.model.state.LoginDialogState
 import com.machikoro.client.domain.model.state.LogoutState
 import com.machikoro.client.domain.model.state.RegisterDialogState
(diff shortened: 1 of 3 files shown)


Moritz2k1 (approved):
Fixes what others have destroyed

PR #130, opened 2026-05-09
feat: keep user on home screen after lobby creation

Ivona-Goranova:
Updated the lobby creation flow on the client side.

Previously, the app navigated directly to the LobbyScreen immediately after a lobby was created. This prevented the generated lobby code from being displayed on the HomeScreen.

Changes:

* Kept the user on the HomeScreen after successful lobby creation
* Displayed the generated lobby code directly below the "Lobby erstellen" button
* Prepared the flow for future navigation via the confirmation/check button instead of automatic screen switching


Ivona-Goranova:
Yes, exactly. This PR is only about keeping the generated lobby code visible on the HomeScreen after creating a lobby.
Entering the lobby screen through the check button will be handled separately in another sub-issue.

eles17 (approved):
Thanks for clarifying the scope. 

Please address the check-button navigation to the lobby screen later in the separate PR / sub-issue as discussed.

PR #131, opened 2026-05-09
Gate lobby/game UI on authenticated state

### LINKED ISSUE #106: Gate lobby/game UI on authenticated state

FabianSolutions:
Closes #106. Server-side dependency #160 already merged.

## Summary

- **HomeScreen** gets an `Abmelden` (Logout) button in the top-left. Wired through `AppRoot` to the existing `LogoutViewModel` + `SessionManager.signOut()`. The unauthenticated start screen already routes via `AppRoot` (`StartScreen` for null session, `HomeScreen` for non-null), so no defensive gating on `StartScreen` itself is needed.
- **STOMP auth-rejection handling**: `OkHttpWebSocketClient` now distinguishes the auth-rejection `ERROR` frame (literal body `"Authentication failed"`, per Server #159's `StompAuthChannelInterceptor`) from generic transport errors. On match it emits a new `authRejections: SharedFlow<Unit>`, flips status to `DISCONNECTED`, and resets game state.
- **MainActivity** collects `authRejections`, calls `SessionManager.signOut()`, and shows a snackbar (`"Sitzung abgelaufen, bitte erneut anmelden"`) via a new `SnackbarHost` on the existing `Scaffold`. Routing then collapses the user back to `StartScreen`.

Strict body equality on the auth-rejection match keeps the failure mode loud if the server contract drifts — better a regressed flow caught by a failing test than a silent fall-through.

## Test plan

Unit + lint:
- [x] `./gradlew :app:testDebugUnitTest` — green
- [x] `./gradlew :app:lintDebug` — no new issues
- [x] `./gradlew :app:assembleDebug :app:assembleDebugAndroidTest` — green

Compose UI tests (compile only — need a device/emulator to run):
- [x] `AppRootTest.showsStartScreenWhenUnauthenticated`
- [x] `AppRootTest.showsHomeScreenWhenAuthenticated`
- [x] `HomeScreenLogoutTest.showsLogoutButton`
- [x] `HomeScreenLogoutTest.clickingLogoutInvokesCallback`

Manual:
- [ ] Cold start → StartScreen (Register/Login/Rules, no Logout, no lobby cards).
- [ ] Login → routes to HomeScreen with Logout + lobby cards.
- [ ] Tap Logout → returns to StartScreen.
- [ ] Auth-rejection: login, background app, wipe DB (`docker compose -f compose-dev.yaml down -v && up -d`), foreground → snackbar appears, UI returns to StartScreen.

## Out of scope (per #106)

- Refresh tokens / silent re-auth.
- Per-action authorization (`NOT_YOUR_TURN` etc.) — server's concern.
- Anonymous-game support.

### DIFF
--- a/app/src/androidTest/java/com/machikoro/client/ui/AppRootTest.kt
+++ b/app/src/androidTest/java/com/machikoro/client/ui/AppRootTest.kt
@@ -19,6 +19,12 @@ import org.junit.Rule
 import org.junit.Test
 
 private const val START_SCREEN_TITLE = "MACHI KORO"
+// HomeScreen-specific labels — distinguish HomeScreen from StartScreen, which
+// shares the "MACHI KORO" title.
+private const val HOME_SCREEN_LOBBY_CARD = "Lobby beitreten"
+private const val HOME_SCREEN_LOGOUT = "Abmelden"
+private const val START_SCREEN_LOGIN = "Login"
+private const val START_SCREEN_REGISTER = "Register"
 
 class AppRootTest {
     @get:Rule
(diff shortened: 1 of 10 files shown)


eles17 (changes requested):
Thanks for the work here. Scope looks aligned with Client #106: auth-gating, logout on HomeScreen, and expired/invalid WS auth handling all belong in this PR.

Before merge, please fix two things:

1. Logout/auth rejection resets game state but does not clear `mutableLobbyCode`, so an old lobby code can reappear after logging in again in the same app session. Please clear lobby state too.

2. `authRejections` uses `MutableSharedFlow(..., replay = 0)`, so the auth rejection can be missed if emitted before `MainActivity` starts collecting. Since this triggers `signOut()`, please make that event durable or handle sign-out somewhere guaranteed to be active.

I also found the network-level auth-rejection tests in `OkHttpWebSocketClientTest`, so that part is covered. The only remaining test gap from #106 is the full sign-out/UI transition path.

eles17 (approved):
Thanks, the requested changes are addressed.
Approve from my side.

Issue #118, opened 2026-05-02
Optional: Card Library


SoupCube:
### User Story
As a player, I want to view all cards available in the game so that I can learn and plan strategies.

### Acceptance Criteria
- A button or menu option opens a card library
- All available cards are displayed
- Each card shows:
  - Name
  - Cost
  - Effect
  - Activation condition
- Cards are organized (e.g. by category or cost)
- View is accessible outside of active gameplay (and optionally during)

### Notes
- Useful for onboarding and strategy
- Could reuse card UI components
- Could use sorting options

#### Related Issues
- https://github.com/SE2-Machi-Koro/Server/issues/182

- https://github.com/SE2-Machi-Koro/Client/issues/114
- https://github.com/SE2-Machi-Koro/Client/issues/111

PR #132, opened 2026-05-11
127 roll dice button restricted to active players turn only

kutschera13:
## Description
The Roll Dice button is now restricted to the active player only. Previously, 
the button was visible and clickable for all players regardless of whose turn it was.

## Acceptance Criteria
- [x] Button is only visible/clickable for the active player
- [x] Button is disabled when it is not the player's turn

## Implementation

**Server:**
- `GameController.broadcastPhase()` now includes `activePlayerId` in every `GAME_ACTION` broadcast

**Client:**
- `LoginResponse` and `Session` now include `userId` so the client knows its own identity
- `OkHttpWebSocketClient` parses `activePlayerId` from `GAME_ACTION` and exposes it as a `StateFlow`
- `GameScreenState` has new fields: `activePlayerId`, `myUserId` and computed `isActivePlayer`
- `GameScreenViewModel` collects `activePlayerId` and `myUserId` from session
- `GameScreen` shows the "Würfeln" button only when `state.isActivePlayer == true`
- `GameScreenViewModel.rollDice()` guards against non-active players calling roll

## Related
- Follows up on #46 (Implement Dice Roll Button)
- Requires Server #118 (broadcast activePlayerId)

## Closes
Closes #127


eles17 (changes requested):
I checked this against Server PR #222: https://github.com/SE2-Machi-Koro/Server/pull/222

The client-side implementation looks good to me, but it depends on the server contract being aligned before the feature works end-to-end.

The client currently expects `activePlayerId` from `GAME_ACTION`, while the server currently sends `activeUserId` in `broadcastPhase()`. Also, the server should include/broadcast the active player id immediately when the game starts, so the client does not enter `ROLL_DICE` before knowing who the active player is.

Suggested project fix: update Server PR #222 to consistently send `activePlayerId`, including on/after `GAME_STARTED`.

With that server change in place, I’m fine approving this client PR.

eles17 (approved):
Approved. I re-checked this against Server PR #222: https://github.com/SE2-Machi-Koro/Server/pull/222

The server-side changes that this client PR depends on are now aligned with the client contract:
- `GAME_ACTION` now uses `activePlayerId`, which matches what the client parses.
- The server now sends a `GAME_ACTION` with `turnPhase` and `activePlayerId` immediately after `GAME_STARTED`, so the active player can see/use the roll button as soon as the game starts.
- `broadcastPhase()` also continues to include `activePlayerId` for later phase changes.

Issue #53, opened 2026-04-13
Add lobby code input and join button


Ivona-Goranova:
## Goal
Allow the player to enter a lobby code and trigger joining.

## Tasks
- [ ] Add text input field for lobby code
- [ ] Add "Join Lobby" button
- [ ] Capture user input

## Definition of Done
- [ ] Input field is visible
- [ ] Button is visible
- [ ] User can enter a code
- [ ] No crashes
- [ ] Tested manually
- [ ] PR created and CI passes

PR #135, opened 2026-05-12
51 navigate to lobby screen

### LINKED ISSUE #51: Navigate to lobby screen
**Goal**
After creating a lobby and displaying the generated lobby code, the player should be able to confirm and proceed to the lobby screen.

**Flow**
- User clicks "Create Lobby" and a unique lobby code is generated by the server and displayed to the user -> #52 
- The user sees a confirmation button 
- After clicking the button, the user is navigated to the Lobby Screen and can see other players

**Tasks**
- Add confirmation button below the code
- Implement navigation to Lobby Screen on button click

Ivona-Goranova:
After creating a lobby and receiving a lobby code from the server, the user can now navigate directly to the LobbyScreen by clicking the confirmation icon.

**Changes**
- Added local navigation state in MainActivity
- Extended AppRoot to switch between HomeScreen and LobbyScreen
- Added onGoToLobbyClick callback support
- Connected the home_check_icon to the lobby navigation flow
- Updated previews and screen parameter handling


eles17 (changes requested):
Merge conflicts have been addressed.

The production routing looks correct now: #51 keeps the user on `HomeScreen` after lobby creation, then navigates to `LobbyScreen` only via the confirm/check icon, and the newer logout wiring from `main` was preserved.

**One blocker before merge:** this changes the required `AppRoot` / `HomeScreen` parameters, but the Android UI tests on `main` were not updated. `compileDebugAndroidTestKotlin` will fail because existing `AppRootTest` calls are missing `showLobbyScreen` and `onGoToLobbyClick`, and `HomeScreenLogoutTest` calls are missing `onGoToLobbyClick`.

This should be fixed in this PR rather than a follow-up, since it is caused by the #51 API change and would leave `main` with non-compiling test sources. The fix should be small: pass `showLobbyScreen = false` and `onGoToLobbyClick = {}` in the existing tests, and optionally add a small AppRoot test for the Home -> Lobby transition.

Also please add:
- one test proving authenticated Home -> Lobby routing when `showLobbyScreen` becomes `true`

eles17 (approved):
Thanks Fabian and Ivona. Approving. 

Please address this small change fabian requested in your next issue #53 
"One small thing worth considering: in HomeScreen, onGoToLobbyClick: () -> Unit is the only callback without a default value — every other callback (onJoinLobbyClick, onCreateLobbyClick, onLogoutClick, etc.) defaults to = {}."

PR #144, opened 2026-05-14
fix: wire HomeScreen Abmelden button to LogoutViewModel

FabianSolutions:
## What's broken

The Abmelden (sign-out) button on the authenticated HomeScreen does nothing. Tapping it has no visible effect — no API call, no session clear, no screen transition.

## Root cause

Three things conspired:

1. `HomeScreen.kt` exposes `onLogoutClick: () -> Unit = {}` with a **default empty lambda** — silent fallback that hides missing wiring.
2. `AppRoot.kt` passes `onLogoutSubmit` only to `StartScreen`, not `HomeScreen`. But routing collapses logged-in users to `HomeScreen` — `StartScreen`'s logout wire is unreachable for them.
3. Issue #106 (the prior PR that placed the Abmelden button on `HomeScreen`) added the parameter but never connected it to the existing `LogoutViewModel.submit`.

`LogoutViewModel.submit()` itself is fine — it calls `authApi.logout(...)` and then `sessionStateHolder.signOut()`. It just had no caller from `HomeScreen`.

## Fix

- `AppRoot.kt`: forward `onLogoutSubmit` to `HomeScreen.onLogoutClick`.
- `HomeScreen.kt`: drop the `= {}` default on `onLogoutClick` so any future missing wire fails at compile time instead of silently no-op'ing at runtime.
- `HomeScreen.kt`: two `@Preview` composables now pass an explicit `onLogoutClick = {}` — they aren't real callers, `{}` is honest there.
- `HomeScreenLogoutTest.kt`: the `showsLogoutButton` smoke test now passes `onLogoutClick = {}` (it only asserts the button is displayed, not click behavior — the second test in the file already wired a real callback).
- `AppRootTest.kt`: pre-existing latent failure surfaced when the now-required `onLogoutClick` forced `compileDebugAndroidTestKotlin` to actually compile this file — backfilled missing `isLobbyHost = false` on two `AppRoot(...)` calls. Pure compile-fix, no behavior change to the test.

## Why tests didn't catch it before

`HomeScreenLogoutTest.clickingLogoutInvokesCallback` passes `onLogoutClick = { clicks += 1 }` directly to `HomeScreen` and asserts `clicks == 1`. It passed because the test bypassed the broken AppRoot wire. The test verified the *component* works in isolation, not that the *production composition* works. After this fix, the silent default lambda is gone, so any future regression of this class fails the build.

## Test plan

- [x] `./gradlew test` green (unit tests).
- [x] `./gradlew compileDebugAndroidTestKotlin` green (androidTest now actually compiles).
- [x] `./gradlew assembleDebug` green.
- [x] After merge, manual smoke: log in on the Android client, tap Abmelden, verify session clears and the start screen reappears.

## Commits

- `3dca8ce` — fix: wire HomeScreen Abmelden button + remove silent default
- `d637072` — test: backfill missing isLobbyHost in AppRootTest (pre-existing latent failure)


Moritz2k1 (approved):
Looks good to me. Wires Abmelden Button to LogoutViewModel

PR #145, opened 2026-05-14
fix: reset Login/Register dialog state when opening

FabianSolutions:
## What's broken

After registering or logging in, returning to the start screen and tapping Login (or Register) shows the dialog **prefilled with the previous attempt's username/password** and a stale 'Logged in as ...' subtitle with only a Close button visible. The user has to manually close and reopen the dialog to get a usable empty form.

Screenshot for reference: dialog shows 'Username: Admin', '••••', and 'Logged in as admin' even after the user has logged out and is back on the start screen.

## Root cause

The dialogs do call `onLoginDialogReset()` / `onRegisterDialogReset()` — but only from their `onDismiss` callback. After a successful submission, the sequence is:

1. `LoginDialogViewModel.submit()` or `RegisterDialogViewModel.submit()` succeeds
2. `sessionStateHolder.signIn(...)` updates `SessionManager.session`
3. `AppRoot`'s `loggedInAs` becomes non-null → routing collapses `StartScreen` to `HomeScreen`
4. `StartScreen` unmounts before the user can dismiss the dialog
5. `onDismiss` never fires, so neither reset is ever called

`LoginDialogViewModel` and `RegisterDialogViewModel` are Activity-scoped, so their state survives the StartScreen unmount. When the user returns to `StartScreen` (after logout) and reopens the dialog, it renders the leftover state — `username = \"admin\"`, `loggedInAs = \"admin\"`, etc.

## Fix

Reset the dialog state on the click that opens it, not just on dismissal:

```kotlin
SecondaryActionButton(
    text = \"Login\",
    onClick = {
        onLoginDialogReset()
        showLoginDialog = true
    },
)
SecondaryActionButton(
    text = \"Register\",
    onClick = {
        onRegisterDialogReset()
        showRegisterDialog = true
    },
)
```

This guarantees the dialog always opens clean regardless of how the previous interaction ended (success-routed-away, dismissed, or never closed). A short comment explains the route-flips-on-success scenario at the call site.

## Test plan

- [x] `./gradlew test` green locally (unit tests).
- [x] `./gradlew compileDebugAndroidTestKotlin` green (androidTest compiles).
- [x] After merge, manual smoke: log in successfully, then log out, then tap Login again — verify the dialog opens with empty fields and no stale 'Logged in as ...' subtitle. Repeat for Register.

## Commits

- `53a49fb` — fix: reset Login and Register dialog state when opening
- `ec15926` — test: backfill missing isLobbyHost in AppRootTest (pre-existing latent failure; same fix as on PR #144, included here so CI can run this PR independently)

## Related

- PR #144 — fixes the Abmelden (logout) button wiring. That PR + this one are independent; either can merge first.


Moritz2k1 (approved):
Extends AppRootTest and extends StartScreen functionality

Issue #55, opened 2026-04-13
Handle success and error response


Ivona-Goranova:
## Goal
Handle backend response for joining a lobby.

## Tasks
- [ ] Navigate to lobby screen on success
- [ ] Display error message for invalid code

## Definition of Done
- [ ] Successful join navigates correctly
- [ ] Invalid code shows error message
- [ ] No crashes
- [ ] Tested manually
- [ ] Works with WebSocket
- [ ] PR created and CI passes

PR #147, opened 2026-05-14
142 fix lobby navigation after lobby code confirmation

Ivona-Goranova:
This PR fixes the lobby navigation flow after lobby creation and confirmation #142 

**Implemented:**
-  The host remains on HomeScreen after creating a lobby.
-  The generated lobby code is displayed on HomeScreen.
- The lobby code can be copied to the clipboard.
- The host navigates to LobbyScreen only after confirming via the check action.
- LobbyScreen also displays the lobby code with copy functionality.
- AppRoot navigation logic and related tests were updated.
- UI was changed from German to English.
- Android and unit tests are passing again.

**Note:** The "_Other players join a lobby_" flow is intentionally not part of this PR and will be handled separately in: #53 and #55

### DIFF
--- a/app/src/androidTest/java/com/machikoro/client/ExampleInstrumentedTest.kt
+++ b/app/src/androidTest/java/com/machikoro/client/ExampleInstrumentedTest.kt
@@ -42,7 +42,7 @@ class ExampleInstrumentedTest {
         }
 
         composeTestRule.onNodeWithText("MACHI KORO").assertIsDisplayed()
-        composeTestRule.onNodeWithText("Connection status: connected").assertIsDisplayed()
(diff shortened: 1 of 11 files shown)


eles17 (changes requested):
Could we adjust `HomeViewModel.createLobby()` so the first `Create Lobby` click continues after the websocket connects?

Right now, when `connectionStatus != CONNECTED`, it calls `connect()` and returns, so the user may need to click `Create Lobby` a second time. Ideally, we should wait for `CONNECTED` and then call `sendCreateLobby()`, while also handling `ERROR` / `DISCONNECTED` so the pending job does not stay active forever.

eles17 (approved):
Thanks for changes. Approved

PR #150, opened 2026-05-15
feat: render full reconnect snapshot from /app/game.sync

FabianSolutions:
Closes #140. Final piece of the SE2-Machi-Koro/Server#246 reconnect chain on the client side.

## What's broken today

The client subscribes only to `/topic/public` and `/topic/game/{id}` — **never to `/user/queue/game-sync`**. The server pushes a full reconnect snapshot there (automatically, after `chat.addUser` detects an active in-progress game), but the client silently drops it. Reconnecting mid-game restores nothing.

### Snapshot parsing (commit `ee029f7`)
- Subscribe to `/user/queue/game-sync` on STOMP CONNECTED — *before* the JOIN send, so the subscription is registered when the server emits the auto-sync.
- Parse the `SYNC` message: it wraps a full `GameStateDto` under `payload.state`. One snapshot restores game status, turn phase, round, players/coins, active player, dice, **landmarks**, and **marketplace**.
- New client enums (`GameStatus`, `CardType`, `LandmarkType`) + `PlayerLandmarkState` model for type-safe deserialization.
- New `WebSocketClient` flows: `gameStatus`, `roundNumber`, `playerLandmarks`, `marketplace`.
- Active player's `userId` resolved from `turnOrder[currentTurnIndex]` → the matching player row (turnOrder holds playerIds).

### Rendering + routing (commit `d20e647`)
- `GameScreen` renders each player's four landmarks as built/unbuilt pips inside their badge, a horizontally-scrollable marketplace supply section, and a round-number indicator.
- `GameScreenViewModel` collects the new flows into `GameScreenState`.
- `AppRoot` routes `gameStatus == FINISHED` to the (previously unwired) `GameOverOneWinner` screen. The winner is derived client-side from the snapshot — the player whose landmarks are all built (Machi Koro's win condition) — so no extra server field is needed.

## Test plan

- [x] `./gradlew test compileDebugAndroidTestKotlin assembleDebug` — BUILD SUCCESSFUL; all unit tests pass, androidTest compiles, APK builds.
- [x] 8 SYNC-parsing tests in `OkHttpWebSocketClientTest` (status/phase/round, players, active-player resolution, dice, marketplace, landmarks, malformed payload, disconnect-reset, sync-queue subscription).
- [x] 4 viewmodel-mapping tests in `GameScreenViewModelTest`.
- [x] 4 Compose rendering tests in `GameScreenSnapshotTest` (round, marketplace, landmark build-state, dice) + 1 `AppRootTest` FINISHED-routing test — instrumented; run on-device.
- [ ] Manual: kill the app mid-game, relaunch → board reconstructs from the snapshot.

## Notes

- **Forward-compatible** with the server concurrently extending `GameStateDto` (e.g. card/landmark definitions) — `handleSync` reads only the fields it needs and ignores extras.
- `lastDiceRoll` — the server snapshot persists only the dice *total*, so on reconnect the dice display shows the total as a single value. Live rolls still show individual dice.
- FINISHED routing fires on the reconnect snapshot (per #140 scope). The live "game just ended" transition needs a separate `GAME_WON` broadcast handler — worth a follow-up issue.

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -108,7 +108,16 @@ tasks.register<JacocoReport>("jacocoTestReport") {
         html.required.set(true)
     }
 
-    val debugTree = fileTree("${project.layout.buildDirectory.get()}/tmp/kotlin-classes/debug") {
+    // AGP's built-in Kotlin compiler emits debug classes under
+    // intermediates/built_in_kotlinc/...; older toolchains used
+    // tmp/kotlin-classes/debug. Include both so JaCoCo finds the classes
+    // regardless of toolchain — otherwise the coverage report is empty and
+    // SonarCloud sees 0% coverage.
+    val debugTree = fileTree(project.layout.buildDirectory.get()) {
+        include(
+            "tmp/kotlin-classes/debug/**",
+            "intermediates/built_in_kotlinc/debug/compileDebugKotlin/classes/**",
+        )
         exclude(coverageExclusions)
     }
     val mainSrc = "${project.projectDir}/src/main/java"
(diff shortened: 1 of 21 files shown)


Moritz2k1 (approved):
Mirrors enums from Server. Implements tests. Implements Snapshot fields and states. Extends Websockets

PR #169, opened 2026-05-18
feat: auto-reconnect the WebSocket after an unexpected drop

FabianSolutions:
## What

`OkHttpWebSocketClient` now automatically reconnects after an unexpected WebSocket drop — a network blip or a backend container restart — using exponential backoff.

## Why

Game-state recovery (disconnect/reconnect and backend container restart) depends on the client re-establishing the WebSocket: on STOMP `CONNECTED` it re-subscribes to `/user/queue/game-sync` and re-sends `chat.addUser`, which makes the server push the recovery snapshot. Previously the client never reconnected on its own — recovery only happened if the user manually backgrounded and re-foregrounded the app, so a backend restart mid-game left every connected player stuck.

## Changes

- On a non-deliberate close/failure with a live session, retry `connect()` with exponential backoff (1s, 2s, 4s, 8s, 16s; the last value repeats).
- Stop retrying on a client-initiated `disconnect()` (including app backgrounding) and on an auth rejection.
- A successful STOMP `CONNECT` cancels any pending retry and resets the backoff.
- The backoff schedule and the coroutine scope are constructor-injectable, so the behaviour is unit-testable.

## Testing

- 6 new unit tests in `OkHttpWebSocketClientTest`: reconnect on failure, reconnect on close, no reconnect on client-initiated disconnect, no reconnect on auth rejection, keeps retrying while the backend stays down, and re-subscribes to the game-sync queue after reconnecting.
- All 94 tests in `OkHttpWebSocketClientTest` pass; `lintDebug` is clean.

## Note

Fully automatic end-to-end recovery also depends on #163 — this change restores the game state on reconnect, but routing to the Game screen still goes through `NavigationViewModel`, which has a separate dropped-navigation bug.

Closes #166

### DIFF
--- a/app/src/main/java/com/machikoro/client/network/websocket/OkHttpWebSocketClient.kt
+++ b/app/src/main/java/com/machikoro/client/network/websocket/OkHttpWebSocketClient.kt
@@ -14,13 +14,19 @@ import com.machikoro.client.domain.model.state.PlayerCoinState
 import com.machikoro.client.domain.model.state.PlayerLandmarkState
 import com.machikoro.client.domain.session.SessionStateHolder
 import java.net.URI
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.Job
+import kotlinx.coroutines.SupervisorJob
 import kotlinx.coroutines.channels.BufferOverflow
+import kotlinx.coroutines.delay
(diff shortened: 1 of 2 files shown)


Moritz2k1 (approved):
Adds auto reconnect logic to frontend. Looks clean to me. Thanks for not deleting the comments

PR #172, opened 2026-05-18
fix: handle session expiration by signing out and navigating to Main …

valentinaschiavon99:
## Summary

Fixes #171.

This PR handles stale persisted sessions that are rejected by the server on WebSocket authentication. When an auth rejection is received, the app now clears the session, resets lobby navigation state, clears the lobby code, and falls back to the Start screen instead of leaving the user stuck on Home/Lobby with non-functional actions.

## Changes

- Clear the active session when `authRejections` is collected in `MainActivity`.
- Reset lobby navigation state and lobby code after auth rejection.
- Make unauthenticated navigation always route to `Main`, even if stale lobby/game state is still present.
- Add regression tests for unauthenticated routing with stale lobby and stale game state.


Moritz2k1 (approved):
Looks good to me.

PR #174, opened 2026-05-18
Fix/173 fix lobby client interaction

Moritz2k1:
Implemented handling for Lobby Rosters, PR on the server should be merged first (#289 on the server side)


eles17 (approved):
Looks good. Approved

PR #183, opened 2026-05-19
Fix/182 presentation bugs

Moritz2k1:
This PR fixes the following things:

- When the app starts, it always, always goes to the StartScreen where the User must register/login
- After that we always, literally always go to the HomeScreen()
- On the top left I have implemented a Purge button to delete all games and dummy players from the database
- When creating a lobby and leaving, the lobby gets deleted if no real people are in there
- After adding dummy players in a lobby, you can also reset the lobby and remove the dummy players
- When being in game, I adjusted the centering of the playerCards and implemented a Leave Button
- Through this button you can leave the game and go back to the HomeScreen
- If the player is in a game, he cannot create a new lobby and must use the new Resume Game button (replaces Public Lobby button)
- If the player has no game to resume, he cannot press the Resume Game button
- Adjusted centering of BottomRow on HomeScreen()

Test the whole thing out yourself and make sure to first merge the Server PR


eles17 (dismissed):
Thanks, the direction makes sense and the client changes line up with the server-side presentation/debug flow. I would not merge this client PR yet, though.

Before merging, Server PR #302 needs to land first. That PR still has a requested change around the debug purge endpoint being enabled/exposed by default.

A couple of client-side issues should also be fixed:

1. `MainActivity` now calls `SessionManager.signOut()` in `onCreate()`. That signs the user out on every Activity recreation, including rotation/config changes or process recreation, not only on a fresh app start. This can kick a logged-in player back to auth mid-flow.

2. The `Purge DB` action is visible unconditionally on `HomeScreen`, and the client clears local game/lobby state after `debugApi.purge()` without checking whether the server actually accepted the request. If the server debug endpoint is disabled or fails, the UI can still pretend the purge succeeded.

So from my side: not ready to merge yet. Please merge/fix the server PR first as you stated and address the two client issues above.

eles17 (approved):
Requested changes are addressed. Approved!

PR #204, opened 2026-05-22
185 block game input until game is in progress

SoupCube:
### Summary
This PR updated the GameScreen to only allow player actions when the Game Status is IN_PROGRESS

### Changes
- Roll dice now also checks for the correct game state
- Purchase buttons check for correct game state
- UI shows an overlay when the game is not ready for input
- ViewModel ignores dice roll and purchase triggers when the gamestate is not IN_PROGRESS
- Tests have been modifeid to account for these changes and additions


FabianSolutions (approved):
Approving. the core change is correct and tested.

PR #206, opened 2026-05-23
Added Tests to ensure correct resource display

SoupCube:
Since the logic already was there, I just added some tests to make sure the syncronisation and display are handled correctly.


eles17 (dismissed):
This PR adds useful tests, but it must satisfy #187 independently against `main`.

The server already provides the complete initial game state in the `GAME_STARTED` payload, including players/coins, landmarks, marketplace data, and shop definitions. The current tests manually emit these values into the ViewModel, so they do not verify that the client actually parses and applies the real server payload.

Please add or verify the production parsing for the initial `GAME_STARTED` message and add a WebSocket parsing test using the actual server payload shape before this PR is used to close #187.

The existing `SYNC` coverage is relevant for reconnect behavior, but it does not replace coverage of the initial game-start flow required by this issue.

valentinaschiavon99 (dismissed):
Thanks for adding more coverage around initial game resource display. I checked this against #187 and the current `main` branch, and I do not think this PR is ready to merge as issue-closing work yet.

Blocking issues:

1. `connectedDebugAndroidTest` does not compile. `app/src/androidTest/java/com/machikoro/client/ui/game/GameScreenSnapshotTest.kt` imports `GameScreen` from `com.machikoro.client.ui.game.ui.GameScreen`, but the composable is declared in `com.machikoro.client.ui.game.GameScreen`. Local result: `:app:compileDebugAndroidTestKotlin FAILED` with `Unresolved reference 'GameScreen'`.

2. Some new UI assertions do not match the current UI contract. The current player is rendered as `You`, not their raw display name, and the visible coin text is the numeric amount rather than `"4 coins"`. Those assertions should be aligned with the current semantics/content descriptions or the UI should be intentionally changed.

3. The marketplace/card availability criterion from #187 is still not demonstrated in the game UI. `MarketplaceSection` is currently guarded by `if (false)` in `GameScreen.kt`, so the stock chips asserted by the snapshot tests are not actually rendered. This means #187's marketplace availability UI requirement is not fully met.

What passed:

- `./gradlew build` passes locally.
- GitHub reports `Client CI` and `SonarCloud Analysis` successful for head `199b175`.
- The WebSocket parser coverage is much stronger now, especially around full `GAME_STARTED` payload parsing and reconnect `SYNC` snapshots.

Please fix the Android test import/assertions, decide whether marketplace stock should be rendered in `GameScreen`, rerun `connectedDebugAndroidTest`, and then request another review. This can be merged as incremental coverage only after the Android test target is clean; it should not close #187 until the marketplace UI criterion is handled.

valentinaschiavon99 (approved):
Approved.

I rechecked the current head against issue #187 and the current main/server contract. The previous blockers are addressed: Android test sources now compile, the marketplace state is rendered in `GameScreen`, and the WebSocket coverage verifies full `GAME_STARTED`/snapshot parsing for players, coins, landmarks, marketplace, and shop definitions.

Validation:
- `./gradlew build` passes locally.
- Current GitHub `Client CI` passes.
- Current SonarCloud analysis passes.
- `./gradlew connectedDebugAndroidTest` compiles the Android test target and only fails locally because no device is connected.

Server compatibility looks correct against current `SE2-Machi-Koro/Server` main.

PR #207, opened 2026-05-23
177 fix apply authoritative game state snapshots from game action broadcasts

eles17:
### Summary

- Apply authoritative server game-state snapshots from GAME_ACTION, ROLL_DICE, GAME_END, GAME_STARTED, and SYNC websocket messages.
- Add player-owned card snapshot state so client state can track playerCards from the backend.
- Align active-player handling with current server websocket contract, where activePlayerId and turnOrder use user IDs.
- Keep purchase success feedback working while reconciling UI state from server snapshots.

### Notes
Plain ./gradlew testDebugUnitTest depends on gradle.properties; current local properties point to the deployed AAU server, while AppConfigTest expects default local URLs.


valentinaschiavon99 (approved):
**No blocking issues found. The PR is mergeable from a client-side perspective.**

The implementation correctly applies authoritative payload.state snapshots from GAME_ACTION, ROLL_DICE, and GAME_END broadcasts, while preserving existing SYNC behavior and purchase success feedback. The active player handling is aligned with the backend contract: activePlayerId and turnOrder are treated as user IDs, which matches the current server GameStateDto.

Backend compatibility looks good, provided the server version containing the real-time snapshot broadcast changes from SE2-Machi-Koro/Server#293 is deployed. No additional backend changes appear necessary for this PR.

**Non-blocking note:** playerCards are now parsed into client state, but they are not currently rendered in the game UI. If visible owned-card display is expected, that should be handled in a follow-up PR.

PR #211, opened 2026-05-24
188 navigate from lobby to game after initialization

SoupCube:
### Changes
Navigation checks State for navigation decision making.
Tests test for state based navigation.


eles17 (changes requested):
Requesting changes

There is one issue before this can be merged:
When a user newly enters a lobby, `gameStatus` is initially `null` until the first server snapshot arrives. With the new condition, the app navigates back to the Home screen instead of keeping the user in the Lobby during this loading state.

Please allow the Lobby route while `gameStatus` is `null` or `WAITING`, and add/restore a test for `showLobby()` with `GameScreenState.initial()`.

The `IN_PROGRESS -> Game` navigation and passing of the `gameId` otherwise look correct.

eles17 (approved):
Approved. Now safe to merge

PR #212, opened 2026-05-25
209 update app theme based on new figma design

Ivona-Goranova:
This PR updates the client UI theme and visual design based on the new Figma mockups.

**Changes:**

- Added and updated the new color palette in the theme package
- Integrated the new font styling for titles 
- Updated multiple drawable assets according to the latest design
- Replaced and cleaned up card and landmark image resources
- Added PNG assets for all dice perspectives


eles17 (changes requested):
The new card_business_center.png asset is added in this PR, but it is not mapped in ShopImageResolver. Since ShopCatalog already uses imageKey = "card_business_center", the Business Center card still falls back to the bakery image instead of showing its new artwork.

Please add the missing mapping:
"card_business_center" -> R.drawable.card_business_center

Also remove the card_business_center exception from ShopImageResolverTest, so the test verifies that every available catalog card resolves to its own drawable rather than the fallback.

eles17 (approved):
Thanks for the changes

PR #216, opened 2026-05-25
fix: align reconnect active player parsing with turn order contract

eles17:
### Summary

- Align reconnect active-player resolution with the backend GameStateDto contract.
- Treat turnOrder entries as user IDs when restoring state from SYNC.
- Prevent numeric collisions between database player IDs and user IDs from assigning active-player controls to the wrong user.
- Update reconnect test snapshots to use backend-compatible turnOrder values.
- Add regression coverage for the ID-collision scenario.

### Problem
During reconnect, the client restores the active player from a server snapshot. When activePlayerId is unavailable, the client falls back to turnOrder[currentTurnIndex].

The backend sends turnOrder as user IDs, but the client previously attempted to reinterpret the selected value as a possible database player ID and convert it again to a user ID.

This could produce the wrong active player when the two ID spaces contain the same numeric value.

### Example:

players:
  player id=202, userId=101
  player id=22,  userId=202

turnOrder: [101, 202]
currentTurnIndex: 1
The active user should be 202, but the previous fallback could match database player ID 202 and incorrectly resolve the active user as 101.

### Changes
Simplified resolveActiveUserId in OkHttpWebSocketClient:

- Use activePlayerId when supplied by the snapshot.
- Otherwise, return turnOrder[currentTurnIndex] directly as the active user ID.
- Removed legacy player-ID conversion logic that conflicts with the current server contract.
- Updated existing SYNC fixtures so turnOrder contains user IDs.
- Added a regression test verifying correct active-player restoration when a database player ID collides with another player's user ID.

### Result
After reconnect:

- The correct active player is restored from the server snapshot.
- Active-player UI controls remain restricted to the correct logged-in user.
- turnOrder handling matches the backend contract.

Related Issue
Closes #178

### DIFF
--- a/app/src/main/java/com/machikoro/client/network/websocket/OkHttpWebSocketClient.kt
+++ b/app/src/main/java/com/machikoro/client/network/websocket/OkHttpWebSocketClient.kt
@@ -773,11 +773,10 @@ class OkHttpWebSocketClient(
      * `activePlayerId`. Current server snapshots already send `activePlayerId`
      * and `turnOrder` in user-id space, so that value is authoritative.
      *
-     * The fallback below exists for older/incomplete snapshots that only have
-     * `currentTurnIndex` + `turnOrder`: if the turn entry matches a player row
-     * id, convert it to that row's userId; otherwise keep the turn entry as the
-     * userId. This preserves old reconnect payloads without breaking the new
-     * server contract.
+     * When `activePlayerId` is absent, `currentTurnIndex` selects the active
+     * user ID directly from `turnOrder`, as defined by the server contract.
+     * It must not be reinterpreted as a database player ID: those independent
+     * ID spaces can contain the same numeric value.
      */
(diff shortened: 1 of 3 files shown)


valentinaschiavon99 (dismissed):
The production fix looks correct and the focused tests pass locally. Please adjust the new regression fixture before merge: when snapshot(firstPlayerId = 202) is used, the same snapshot still keys playerCards and playerUsernames with "11". That makes the payload internally inconsistent with the backend DTO shape, because the first player database id is no longer 11.

Please parameterize those fixture keys from firstPlayerId as well, so the ID-collision test remains representative of a real reconnect snapshot.

valentinaschiavon99 (approved):
Rechecked after the fixture update. The collision regression payload is now internally consistent, and the focused websocket tests pass locally. No remaining blocking issues from my side.

PR #228, opened 2026-05-30
fix: Should trigger the debug endpoints now correctly

Moritz2k1:
Fixes the bug that the lobby cannot be filled with dummies.
SessionHeader is passed now


FabianSolutions (approved):
LGTM. Should trigger the debug endpoints now correctly.

PR #235, opened 2026-06-01
feat: debug-only End game button (#191)

FabianSolutions:
## What

Adds a debug-only **"End game"** control on the in-game screen that ends the current game with the local player as winner — the client counterpart to Server #305 (server PR SE2-Machi-Koro/Server#337). Tapping it calls `POST /debug/end-game`; the existing `GAME_END` WebSocket handler then drives the winner screen exactly as a real win would. Closes #191.

## How

- **`DebugApi.endGame(EndGameRequest(gameId))`** (Retrofit) — admin `Authorization: Bearer` header is attached by the existing `DebugApiFactory` interceptor.
- **`GameScreen`** renders a subtle red "End game" text next to "Leave", gated by `BuildConfig.DEBUG && gameStatus == IN_PROGRESS && gameId != null && myUserId != null` — so it's a member-of-a-running-game check and is compiled out of release builds. Mirrors the existing "⚙ Purge DB" debug control.
- **`GameScreenViewModel.endGame()`** posts to the endpoint with the current `gameId`, guards against rapid double-taps (single in-flight call), and on success does nothing — the server's `GAME_END` broadcast is the single source of truth for the transition (no extra client navigation). Failures surface through a `debugEndGameErrors` SharedFlow that `MainActivity` shows as a snackbar.

## Testing

- **Unit tests** (`GameScreenViewModelTest`): call path threads the right `gameId`; no-op when `gameId` is null; in-flight double-tap guard; non-2xx, exception, and null-message error paths each surface the expected snackbar text and emit nothing on success.
- **Lint + coverage:** `./gradlew check` green (Android Lint clean, jacoco gate passes).
- **Manual e2e (2 emulators + local backend):** in a 2-player game, tapping the button navigated **both** clients to the winner screen with the tapper as winner; backend logged the win + cleanup with no errors and no client-side crashes.

## Notes

- Requires Server #305 (PR #337) for the endpoint. Debug-only; ends the game for everyone immediately and is absent in release builds.
- While testing this, I found the lobby "ready/not ready" status is a pre-existing client-only placeholder (not synced across clients) — filed separately as #231; unrelated to this change.


Moritz2k1 (approved):
Implements a quick way to end the game. Can only be triggered by the admin since this is a debug endpoint. Looks fine code wise

PR #241, opened 2026-06-02
210 redesign homescreen

Ivona-Goranova:
**Changes**
- redesigned and polished the HomeScreen to match the Figma sketches
- redesigned the bottom menu layout and styling
- improved welcome label positioning and spacing
- updated background styling
- redesigned the main action cards with custom shadows and decorative screw assets


doniliks (dismissed):
i personally dont like updated bottom bar with buttons, previous was better idk. Also now text is not centered in button which looks messy.
You didnt remove rules from start screen and didnt move them to rules at home screen. 
Also changed color for secondary action button shadow not really suits. 
Field to entry lobby code is now not centered and not width sized to button itself. 
And please return background with cards and coins on there since now it looks empty

eles17 (commented):
Thanks for the update. I checked the latest branch and most of the requested changes are still valid:

- Rules is added to the HomeScreen, but it is still present on the StartScreen, so it was not fully moved.
- HomeScreen now uses `background_wood` instead of `home_screen_background`, so the cards/coins background is gone.
- Bottom bar item content is start-aligned inside fixed-width button cards, so the text is not centered.
- Join lobby code input is centered, but it is still narrower than the Join Lobby card.
- `SecondaryActionButton` shadow color was changed globally from `ButtonBeigeDark` to `ButtonBorderBeige`, which affects more than this screen.

The branch compiles successfully with `./gradlew compileDebugKotlin`, so these are design/UX issues rather than build failures.

eles17 (approved):
Merging this based on the grounds to keep the flow going. If it's functional for now it's fine for me. Rest will be addressed in further issues. 
Other details and personal opinions about design can be addressed later on.

PR #243, opened 2026-06-02
test: add two-player websocket synchronization coverage

eles17:
### Summary

- Added two-client WebSocket synchronization coverage for issue #180.
- Verified non-sender clients update from authoritative broadcasts for GAME_STARTED, GAME_ACTION, ROLL_DICE, purchase/build updates, and reconnect SYNC.
- Replaced embedded literal NUL bytes in OkHttpWebSocketClientTest with \u0000 escapes so the test file is searchable as text.

### Testing
./gradlew testDebugUnitTest
./gradlew test

### Notes
Checked #181 protocol docs against the current implementation; no doc changes were needed.


valentinaschiavon99 (dismissed):
Request changes.

The new two-client websocket coverage is a useful addition, but two tests currently model behavior that does not match the backend contract:

1. `twoClientsApplyGameActionBroadcastToNonSender` sends `/app/game.advancePhase`. The backend deliberately rejects this deprecated endpoint with `DIRECT_PHASE_ADVANCE_FORBIDDEN`. Please use a supported gameplay action such as `resolveEffects` or `endTurn`.

2. `twoClientsApplyRollDiceBroadcast` expects the roll-dice snapshot to transition directly to `BUY_OR_BUILD`. The backend flow is `ROLL_DICE -> RESOLVE_EFFECTS -> BUY_OR_BUILD`, so the expected phase after rolling should be `RESOLVE_EFFECTS`.

No backend changes are required. Once the fixtures reflect the existing server behavior, this should be ready to merge.

valentinaschiavon99 (dismissed):
Requesting changes: the new two-client harness delivers game-topic messages directly to both WebSocket listeners before either client has joined a lobby or subscribed to /topic/game/{gameId}. This means the tests can pass even when the real join/subscription flow is broken.

Please initialize both clients through realistic lobby messages first, assert that both subscriptions exist before broadcasting GAME_STARTED, and make broadcastToBoth deliver only to subscribed clients. The test suite and repository check task otherwise pass.

valentinaschiavon99 (approved):
Approved. I re-checked the current head against the earlier review concerns: the two-client harness now establishes realistic lobby/game-topic subscriptions before game broadcasts, broadcasts only to subscribed clients, uses the supported /app/game.resolveEffects action instead of the deprecated direct phase-advance endpoint, and expects the roll flow to enter RESOLVE_EFFECTS before BUY_OR_BUILD.

PR #246, opened 2026-06-03
242 adjust phase displaying

doniliks:
## Changes
- phase banner is now align to sketches
- current phase is indicated correctly based on state phase value

- added custom text field to show hints to player during the game e.g. "Choose how many dices to roll". This has to be adjusted after server-side for phase advance works correctly.

- further issues will be to add those custon hints and make a proper reset


Moritz2k1 (approved):
LGTM

PR #247, opened 2026-06-04
fix: reset dice count selection to 1 at the start of each turn

kutschera13:
Implemented in branch `197-pass-selected-dice-count-to-rolldice-on-würfeln-press`.

- `onRollDice(diceCount: Int)` in `GameScreen.kt` passes the selected dice count to the ViewModel
- `GameScreenViewModel.rollDice(diceCount: Int = 1)` forwards the count to `webSocketClient.rollDice(diceCount)`
- If the train station is not built, the dice selection UI is hidden and `diceCount` defaults to `1`
- Dice count selection resets to `1` at the start of each turn via `remember(state.roundNumber)`

All acceptance criteria are met.


eles17 (commented):
Small robustness suggestion: please make the no-Train-Station fallback explicit when invoking `onRollDice`.

The current behavior works because the selector is hidden and `selectedDiceCount` resets to `1`, but that couples the game rule to UI state. Since the issue explicitly requires “without Train Station, always send `diceCount = 1`”, it would be clearer and safer to clamp at the action boundary:

`onClick = { onRollDice(if (state.hasTrainStation) selectedDiceCount else 1) }`

This still keeps the Train Station behavior correct: each turn starts on 1 die, and the player can actively choose 2 before rolling.

eles17 (approved):
Thanks. Approved

PR #253, opened 2026-06-04
252 adjust game screen layout

doniliks:
Screen was divided into parts to ensure game content is surrounded by other UI such as top bar and additional buttons on right/left sides.


eles17 (commented):
The layout split into `GameScreenLayout` makes sense for issue #252, but I think a couple of behavior/layout regressions should be fixed before merging:

1. `GamePhaseBanner` no longer receives `state.customDisplayText`. This was passed before the refactor and should stay, otherwise any custom phase prompt text is silently lost.

2. The `End turn` / `Resolve effects` action button moved out of the bottom controls column. The callback still looks correct, but the button is now placed as a direct child of the center `Box`, so it may appear in the wrong position instead of staying grouped with the dice/result controls.

Also, the `gradle/verification-metadata.xml` changes look unrelated to the UI layout task. Please revert them unless they were intentional and needed.

The main button callbacks themselves look correctly wired.

Moritz2k1 (approved):
LGTM

PR #258, opened 2026-06-04
Add "Insider tip" button to trigger the Insider Trading cheat (#255)

FabianSolutions:
## Summary

Adds a **DEBUG-only "Insider tip" button** to the in-game "Leave Game?" dialog that triggers the **Insider Trading** cheat (#203) — the same effect as physically shaking the device. Shaking is impossible on emulators and awkward on tablets / for players with motor impairments (an open question raised in #203), so this gives a reliable, non-physical way to fire the recommendation.

Client-only change. Reuses the existing local recommendation path — no new logic, no Server changes, no WebSocket traffic.

Closes #255.

## What changed

- **`GameScreen.kt`** — the Leave dialog now hosts a `Debug` section (compiled out of release builds via `BuildConfig.DEBUG`) containing the **Insider tip** trigger and the existing **End game** control. Tapping "Insider tip" calls the same `onShake()` the accelerometer uses, so behaviour is identical to a real shake — including the silent no-op when it isn't your turn (the guard lives in `GameScreenViewModel.onShake()`).

Done in two commits:
1. `feat` — add the `Insider tip` button.
2. `refactor` — move the debug actions (Insider tip + End game) out of the dialog's button row into a stacked `Debug` section in the body. With three buttons the row overflowed and "Insider tip" wrapped/overlapped "Stay"; the new layout keeps `Leave`/`Stay` as the only action buttons and prevents any wrap/clip.

## Testing

- `./gradlew :app:assembleDebug :app:lintDebug` — builds; **Android Lint: 0 issues**.
- Manually verified end-to-end on an emulator in a **real 2-player game**: in the Buy phase as the active player, opened the Leave dialog → tapped **Insider tip** → the recommended card (Convenience Store, highest single-roll EV at 3 coins) gets the green highlight — exactly as a physical shake does. Re-verified after the layout refactor.
- The cheat's existing pure logic (`recommendBestBuy`, `onShake`) is already unit-tested; `ui/` is excluded from coverage by project convention, so no new unit test was added.

## Notes

- DEBUG-gated only — the cheat stays hidden in release builds, preserving its intent.
- No Server impact; the recommendation is computed locally from the existing `GameScreenState` snapshot.

### DIFF
--- a/.github/workflows/sonarcloud.yml
+++ b/.github/workflows/sonarcloud.yml
@@ -39,7 +39,7 @@ jobs:
         run: chmod +x gradlew
 
       - name: Build, run tests, generate JaCoCo report and scan with SonarCloud
-        run: ./gradlew sonar --no-daemon --stacktrace
+        run: ./gradlew sonar --no-daemon --stacktrace -Dsonar.scanner.skipJreProvisioning=true
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           SONAR_TOKEN: ${{ secrets.SONAR_TOKEN }}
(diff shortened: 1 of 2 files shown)


doniliks (approved):
Works, merging

PR #262, opened 2026-06-05
Feat 119 dynamic host start button

### LINKED ISSUE #119: One small gap to track: nothing in production sets state.isHost = true yet, so the host button still doesn't render in the running app — only via test stubs. Pairs naturally with Server #165's start-game flow when that lands.
**Description**
Currently, the UI logic for rendering the "Host" button is implemented, but the state.isHost flag is never dynamically set in the production environment. As a result, the host-specific controls only appear when using test stubs and remain hidden in the live application.

This issue serves as a follow-up to the initial UI implementation to ensure the frontend correctly identifies the host based on real data coming from the backend.

**Context**
This task is blocked by or "pairs" with Server Issue #165. Once the server begins broadcasting the host_id (or equivalent) within the lobby state, the frontend must be updated to compare this ID with the current user's ID.

**Proposed Changes**
Data Mapping: Update the lobby state listener/provider to extract the host information.

**State Logic:** Implement the logic to set StartScreenState.isHost = true if currentUser.id === lobby.hostId.

**Verification:** Ensure the "Start Game" button renders correctly for the host and remains hidden for guest players.

**Acceptance Criteria**
[ ] isHost is no longer hardcoded or reliant solely on stubs.

[ ] The host button renders in the production build when the user is the lobby creator.

[ ] The state updates reactively if the host leaves and a new host is assigned (if applicable).

**Dependencies**
Requires Server #165 (Start-game flow/Host ID broadcast) to be merged.

valentinaschiavon99:
Summary
Derive lobby host ownership from backend-provided host identifiers instead of setting isLobbyHost optimistically when creating a lobby.
Preserve the existing create-lobby auto-join handshake with a private pending-create flag that is independent from host authorization state.
Update host ownership reactively from lobby creation, lobby join, lobby roster, and game snapshot payloads.
Issue
Closes #119

Implementation Notes
sendCreateLobby() no longer marks the local user as host before server confirmation.
isLobbyHost is updated by comparing the current session userId with server host fields such as hostId, host_id, hostUserId, and host_user_id.
LOBBY_ROSTER can also resolve host ownership from roster entries marked with isHost or host.
The host flag is reset with lobby state cleanup to avoid leaking stale permissions across lobbies.
Verification
./gradlew testDebugUnitTest
Dependency Context
This frontend behavior expects the server to broadcast host ownership data in lobby state payloads, as described by the paired server work for host ID broadcast support.

### DIFF
--- a/app/src/androidTest/java/com/machikoro/client/ui/game/GameScreenSnapshotTest.kt
+++ b/app/src/androidTest/java/com/machikoro/client/ui/game/GameScreenSnapshotTest.kt
@@ -1,5 +1,8 @@
 package com.machikoro.client.ui.game
 
+import androidx.compose.runtime.getValue
+import androidx.compose.runtime.mutableStateOf
+import androidx.compose.runtime.setValue
 import androidx.compose.ui.test.assertIsDisplayed
 import androidx.compose.ui.test.junit4.createComposeRule
 import androidx.compose.ui.test.onNodeWithContentDescription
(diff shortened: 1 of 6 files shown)


eles17 (approved):
LGTM. Approved

PR #269, opened 2026-06-06
Fix: Race condition when joining a lobby

Moritz2k1:
Fixes the lobby race condition as described in the issue #268


FabianSolutions (approved):
LGTM

PR #276, opened 2026-06-06
256 display coins and cards of current player on game screen

doniliks:
Displays current player coins based on isCurrentPlayer flag. 

NEW:
Displays cards of current player based on isCurrentPlayer flag, rendered only if player is not required to action right now. 
Added 10 seconds timeout with timer and closes it after that to ensure player misses no information
Screenshot 2026-06-08 at 15 03 22
Screenshot 2026-06-08 at 15 03 30
Screenshot 2026-06-08 at 15 03 37

Closes #256, #249


eles17 (commented):
There’s a crash path in PlayerCoinField: state.players.first { it.isCurrentPlayer } throws when the screen renders before players are loaded, or if no player is marked current yet. GameScreenState.initial() starts with an empty players list, and GameScreen always composes this field.

Could you switch this to firstOrNull { it.isCurrentPlayer } and hide the coin field or show a safe fallback until the current player exists?

eles17 (commented):
Thanks, the `PlayerCoinField` crash path from my last comment is fixed now with `firstOrNull`.

I don’t think this is ready to merge yet though:

1. `GameScreen.kt` still renders a raw debug text overlay with game id, phase, players, dice, purchase state, cards, landmarks, etc. This looks like temporary debug UI and would be visible in the game screen.

2. `PlayerCards.kt` uses `.take(4)` for establishments in both the compact and expanded card displays, so players with more than 4 cards will not see all owned cards.

Could you remove the debug overlay and make sure all current-player cards are displayed before merge?

eles17 (approved):
Looks good now, thanks for the fixes.

PR #278, opened 2026-06-07
277 further game screen adjustments

doniliks:
## Changes
- adjusted content paddings
- added marketplace button placeholder
- align contents based on phase e.g. moved advance phase button to the right and rolled number to the left to display other important UI in the middle
- adjusted activePlayer based view e.g. shop if active, "Wait" if other

- additionally changed shop ui splitting cards into categories with buttons to switch between them

Screenshot 2026-06-07 at 20 25 18

Screenshot 2026-06-07 at 20 27 15


eles17 (commented):
Thanks, the wide landscape screenshots look good.

One layout concern remains in the Landmarks tab: it renders the landmark cards in a plain non-scrollable `Row`, while each card is fixed at `165.dp` wide. This works in the screenshot because only two landmarks are visible, but it can overflow or clip when a player has 3-4 unbuilt landmarks, or on narrower/non-landscape layouts.

Could you make the Landmarks tab scrollable or responsive, similar to the Establishments tab? For example, use a `LazyRow`, `LazyVerticalGrid`, or responsive card sizing so all available landmarks remain reachable.

doniliks:
Works with all 4 landmarks as well
Screenshot 2026-06-07 at 20 42 45

eles17 (approved):
Thanks Danylo. Approving.

PR #283, opened 2026-06-08
fix: preserve local player identity after reconnect sync

### LINKED ISSUE #281: fix: preserve local player identity after WebSocket reconnect sync
After WebSocket reconnect, the client restores game state from the `/user/queue/game-sync` `SYNC` snapshot, but local player identity can be mixed up with the active turn player.

Currently, the reconnect snapshot logic appears to mark `PlayerCoinState.isCurrentPlayer` based on whose turn it is. This means if Bob is active after reconnect, Bob may be shown as “You” even when the logged-in user is Alice.

### Expected behavior:
- `isCurrentPlayer` means the logged-in/local user.
- `isActivePlayer` means whose turn it is.
- After reconnect sync, the “You” marker stays on the local player.
- Turn controls remain enabled only for the actual active player when that active player is also the local user.
- `OkHttpWebSocketClientTest.reconnectingClientRestoresStateFromSyncToMatchConnectedClient` passes.
- `./gradlew test` passes.

### Relevant file:
`app/src/main/java/com/machikoro/client/network/websocket/OkHttpWebSocketClient.kt`

eles17:
## Summary
Fixes local player identity after WebSocket reconnect sync.

Previously, `OkHttpWebSocketClient` used the active turn user id to set both:
- `PlayerCoinState.isCurrentPlayer`
- `PlayerCoinState.isActivePlayer`

That meant after reconnect, the player whose turn it was could incorrectly appear as “You”.

## Changes
- Split player identity resolution in WebSocket snapshot parsing:
  - `isCurrentPlayer` now comes from the logged-in session user id.
  - `isActivePlayer` now comes from the active turn user id resolved from `activePlayerId` or `turnOrder[currentTurnIndex]`.
- Kept `activePlayerId`, WebSocket destinations, payload formats, and public models unchanged.
- Updated reconnect sync tests so shared game state is compared separately from client-local “You” flags.
- Added regression coverage for Alice being the local player while Bob is active.

## Verification
- `./gradlew test`


valentinaschiavon99 (approved):
Approved. The PR-specific change meets issue #281: reconnect SYNC parsing now keeps local identity (`isCurrentPlayer`) separate from active turn identity (`isActivePlayer`), remains compatible with the current server `GameStateDto` user-id contract, and local `./gradlew test` / `./gradlew build` plus GitHub CI and SonarCloud are green.

The Train Station landmark-key mismatch and `connectedDebugAndroidTest` import failure are existing main-branch issues, not regressions introduced by this PR.

PR #284, opened 2026-06-08
fix: separate current and active player coin identity

eles17:
Fixes player coin display mismatch between `PlayersTopBar` and the right-side player/coin UI after resolving effects and switching turns.

### **What Changed**

- Updated WebSocket game-state parsing so player identity flags are derived from the correct IDs:
  - `isCurrentPlayer` now uses the local session user ID.
  - `isActivePlayer` now uses the active turn user ID.
- Kept `PlayerCoinState.id` unchanged as the database/player ID, so existing coin, card, and landmark mappings continue to use the same identifier.
- Added regression test coverage for snapshots where:
  - the local user is not the active player,
  - player database IDs and user IDs can overlap,
  - multiple clients receive the same authoritative state but correctly keep different local `isCurrentPlayer` flags.

### **Why**

Previously, both `isCurrentPlayer` and `isActivePlayer` were derived from the active player ID. This caused the `"You"` marker and current-player display behavior to follow the active turn instead of staying attached to the logged-in user. As a result, the top bar could display/hide/highlight coin information for the wrong player after effects resolved or turns switched.

### **How It Works Now**

- Coin amounts still come from `player.coins`.
- `PlayerCoinState.id` still represents the player/database ID.
- Active turn highlighting follows `activeUserId`.
- Local player identity follows `sessionStateHolder.session.value?.userId`.

### DIFF
--- a/app/src/main/java/com/machikoro/client/network/websocket/OkHttpWebSocketClient.kt
+++ b/app/src/main/java/com/machikoro/client/network/websocket/OkHttpWebSocketClient.kt
@@ -1258,14 +1258,19 @@ class OkHttpWebSocketClient(
         val currentTurnUserId = payload.optJSONArray("turnOrder")
             ?.takeIf { currentTurnIndex != null && currentTurnIndex in 0 until it.length() }
             ?.let { turnOrder -> currentTurnIndex?.let(turnOrder::optInt) }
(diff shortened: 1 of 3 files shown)


doniliks (approved):
Works, thanks

PR #287, opened 2026-06-08
fix: update hasTrainStation logic to use active player's database ID

valentinaschiavon99:
## Summary
- Fixes Train Station detection in `GameScreenState.hasTrainStation`.
- Resolves the active player’s database player ID from `players.firstOrNull { it.isActivePlayer }?.id`.
- Keeps compatibility with current server `main`, where `activePlayerId` is a user ID and `playerLandmarks` is keyed by database player ID.

## Why
The UI previously indexed `playerLandmarks` with `activePlayerId`. This failed when `activePlayerId != playerId`, preventing the active local player from seeing the 1/2 dice selector after building Train Station.

## Testing
- Added unit coverage for mismatched `userId` and `playerId`.
- Verified `hasTrainStation == true` when the active player’s database player ID has a built Train Station.
- Verified `hasTrainStation == false` when the active player does not have a built Train Station.

## Verification
- `./gradlew test`
- `./gradlew build`

Closes #285


eles17 (approved):
Everything seems correctly addressed. Approved 👍

PR #289, opened 2026-06-08
Cheating accusations (client): tap-to-accuse + cheat self-report + result toast (#280)

FabianSolutions:
## Summary

Client side of the **cheating-accusations** feature (#280). Players tap another player to accuse them of using the **Insider Trading** cheat (#203, #255); the server (SE2-Machi-Koro/Server#361, PR SE2-Machi-Koro/Server#363) adjudicates and applies the coin penalty.

## What changed

**1. Silent cheat self-report**
- `GameScreenViewModel.onShake()` now also calls `webSocketClient.reportCheat(gameId)` when the local active player uses the cheat. Both the shake gesture *and* the "Insider tip" button (#255) route through `onShake`, so both report — no extra wiring.

**2. Tap-to-accuse UI**
- `PlayerCoinBadge` in `PlayersTopBar` is now tappable for **other** players (not yourself). Tapping opens a confirm dialog ("Bet that X used the Insider Trading cheat. If you're wrong, you lose a coin.") → `GameScreenViewModel.accuse(accusedPlayerId)` → `/app/game.accuse`.

**3. Result feedback**
- `OkHttpWebSocketClient` handles `ACCUSATION_RESULT`: applies the embedded state snapshot (so coin penalties show up via the normal path) and emits an `AccusationResult` on a new `accusationResults` SharedFlow.
- `MainActivity` toasts the outcome: "X caught Y cheating!" / "X wrongly accused Y".

**Plumbing:** new WS destinations (`reportCheat`, `accuse`) + `WebSocketClient` interface methods + `accusationResults` flow + `AccusationResult` model.

## Wire contract (matches server #361)

- **Send:** `/app/game.reportCheat {gameId}` ; `/app/game.accuse {gameId, accusedPlayerId}` (accusedPlayerId = `PlayerCoinState.id`, a `PlayerModel.id`).
- **Receive:** `ACCUSATION_RESULT`, payload `{outcome, accuserPlayerId, accusedPlayerId, penalizedPlayerId, state}`. Parsed by exact keys (kept in lockstep with the server — the #353 lesson).

## Testing

- ✅ `:app:assembleDebug` + `:app:testDebugUnitTest` green (test doubles updated for the extended interface).
- ⏳ **Live end-to-end** (server + 2 emulators) is pending: a separate **lobby-navigation** issue (the app doesn't open the Lobby screen after create/join against the local latest-server build) blocked the live run; being investigated next. The accusation code itself is unit-verified, and the server half has a full real-Postgres suite.

## Related

- Server half: SE2-Machi-Koro/Server#361 (PR SE2-Machi-Koro/Server#363)
- Insider Trading cheat: #203, #255

Refs #280


eles17 (dismissed):
Thanks for the implementation. Beside the failing tests / low coverage, I found a few contract and acceptance issues that should be fixed before merge:

1. `ACCUSATION_RESULT` parsing is not in lockstep with the accepted client/server contract.

Client issue #280 specifies that the client should parse:

`{ accuserPlayerId, accusedPlayerId, caught, penalizedPlayerId, penaltyCoins }`

but this PR currently parses `payload.outcome == "CAUGHT"`. This matches the current server PR implementation, but it conflicts with the original client/server issue contract and with the review already left on `SE2-Machi-Koro/Server#363`.

If the server is fixed per that review, this client will treat every result as a wrong accusation because `outcome` will be absent. Please align both PRs on the same payload shape.

2. The client does not enforce one accusation per accuser per turn.

`GameScreenViewModel.accuse()` sends every accusation while the game is `IN_PROGRESS`, and the UI allows repeated confirmation clicks across the same turn. Client issue #280 explicitly says to disable a second accusation in the same turn, mirroring the server rule. This is the same acceptance gap called out on the server PR: without both sides enforcing it, a player can repeatedly accuse and trigger repeated penalties.

3. `reportCheat` fires even when Insider Trading produced no recommendation.

Issue #280 says to report cheat usage after `recommendBestBuy(...)` returns non-null. The current implementation calls `reportCheat(gameId)` regardless of whether the recommendation is null. Since `recommendBestBuy()` can return null when nothing affordable/in-stock has positive EV, this can falsely mark a player as catchable even though the cheat did not actually produce a recommendation.

eles17 (commented):
Thanks, I re-checked the latest client changes against the previous review feedback and the matching server PR.

The requested client changes look addressed:
- `ACCUSATION_RESULT` now parses the agreed contract fields: `caught`, `accuserPlayerId`, `accusedPlayerId`, `penalizedPlayerId`, `penaltyCoins`.
- The client now enforces one accusation per turn and disables the UI after an accusation.
- `reportCheat` is only sent when Insider Trading actually produces a recommendation.

I also checked compatibility with `SE2-Machi-Koro/Server#363`; the WebSocket destinations and payload fields now appear aligned.

CI/unit tests are green. Only minor cleanup I noticed is a duplicate doc comment before `handleAccusationResult` in `OkHttpWebSocketClient`.

Before merging, I’d still recommend one emulator end-to-end smoke test with the matching server branch, since this is WebSocket/UI/game-flow behavior and the PR notes live E2E was pending.

Moritz2k1 (approved):
Client side of the cheating/accusation functionality. Since the server pr has been merged, I will also merge this

PR #291, opened 2026-06-08
feat: add cards counter displaying for own cards row

doniliks:
Added basic circle with quantity of current player cards, rendered only if more than 1 card was bought

Screenshot 2026-06-09 at 00 25 55


eles17 (approved):
LGTM very nice 👍

PR #295, opened 2026-06-09
fix: require confirmation before buying shop cards

eles17:
## Summary

- Changed shop card clicks to only select/highlight a card instead of buying immediately.
- Reused the main turn action button as `Buy card` during the buy/build phase.
- Added a `Skip` button below the buy button to end the turn without purchasing.
- Automatically ends the turn after the server confirms a successful purchase.

### Current new buy/build flow:

1. Player enters the `BUY_OR_BUILD` phase.
2. Marketplace/shop cards are shown.
3. Player clicks a card once.
4. The clicked card is selected and visually highlighted.
5. No purchase is sent yet.
6. The main right-side action button now says `Buy card`.
7. `Buy card` is enabled only when a card is selected.
8. Player clicks `Buy card`.
9. Client sends the purchase request for the selected card.
10. While purchase is pending, the selected card stays in pending state and buying is disabled.
11. If the server confirms the purchase, the UI shows the success feedback.
12. The client automatically sends `endTurn`.
13. Turn advances through the normal server-driven turn flow.

### Skip flow:

1. Player enters `BUY_OR_BUILD`.
2. Player does not want to buy anything.
3. Player clicks `Skip` below `Buy card`.
4. Client sends `endTurn`.
5. Turn advances normally.

So the key change is: card click = select only, `Buy card` button = actual purchase.


doniliks (approved):
LGTM very nice

PR #308, opened 2026-06-09
292 refactor landmark cards, home logout button and players top bar

Ivona-Goranova:
- Replaced landmark card assets with higher-quality image exports
- Added frame overlay for selected shop cards
- Redesigned the players top bar
- Renamed logout button
- Updated UI components to use Material Theme colors


doniliks (dismissed):
PlayerCoinField - adjust padding between text and image 
PlayerTopBar - remove orange color for active player, adjust only alpha value for inactive players, means 100% opacity for all players but active player displayed bigger.
Also you have removed functionality for „inspect player cards“, 

Shop - Change box to wrap content size and apply size to card image only, based on it place this border image on card, now it does not adjusts itself automatically to screen size.

PR #313, opened 2026-06-09
fix: centralize card definitions and activation sorting

eles17:
## Summary
- Centralize establishment card metadata into a shared client-side `CardDefinition` model.
- Store activation values as numeric `activationNumbers` and derive `activationText` for display.
- Parse websocket/server `cardDefinitions` into the same canonical model.
- Update shop, marketplace, player-card inventory, and cheat recommendation ordering/metadata usage to rely on numeric card definitions.
- Purchase flows still use server enum names via `CardType.name`.

### DIFF
--- a/app/src/main/java/com/machikoro/client/domain/cheat/InsiderTrading.kt
+++ b/app/src/main/java/com/machikoro/client/domain/cheat/InsiderTrading.kt
@@ -2,7 +2,7 @@ package com.machikoro.client.domain.cheat
 
 import com.machikoro.client.domain.enums.CardType
 import com.machikoro.client.domain.enums.LandmarkType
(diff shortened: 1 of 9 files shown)


valentinaschiavon99 (approved):
Approved.

I reviewed this against issue #298 and found the acceptance criteria covered: card definitions are centralized, activation display text is derived from numeric activation values, sorting now uses numeric activation order, WebSocket `cardDefinitions` are parsed into the same client model, and purchase flows still send valid `CardType.name` values to the server.

I also verified compatibility with current `SE2-Machi-Koro/Server` `main`: the server `CardDefinitionDto` sends `activationNumbers: List<Int>` and the `CardType` enum values match the client.

No blocking bugs, errors, UI regressions, or server-contract incompatibilities found.

PR #316, opened 2026-06-10
feat: centralize REST error parsing with ClientError sealed type

valentinaschiavon99:
## Context

Closes #264

Multiple REST call sites (login, register, and others) each duplicated their own
`Throwable.toUserMessage()` logic to handle `HttpException` and `IOException`.
This produced inconsistent behavior: some callers swallowed JSON error details,
none preserved a typed error code, and coroutine cancellation was not explicitly
protected against being silently swallowed.

### `ClientError.kt` — sealed error type
A sealed class with three variants:
- `Http(statusCode, serverErrorCode, message)` — for HTTP errors; carries the raw
  status code, an optional structured server error code (from JSON), and the
  user-facing message.
- `Network(message)` — for `IOException` (no connectivity, timeout, etc.)
- `Unknown(message)` — for anything else that isn't an HTTP or network failure.

### `RetrofitErrorParser.kt` — centralized `Throwable.toClientError()`
A single extension function that all call sites can use:
- **JSON error body** `{"errorCode":"…","message":"…"}` → `ClientError.Http` with
  both fields populated. `ignoreUnknownKeys = true` so future server fields don't break parsing.
- **Plain-text error body** → used verbatim as `message`; `serverErrorCode` is `null`.
- **Empty / blank body** → falls back to `"<fallbackMessage> (HTTP <code>)"`.
- **`IOException`** → `ClientError.Network` with the exception message (or a default).
- **Any other `Throwable`** → `ClientError.Unknown`.
- **`CancellationException`** → always re-thrown, never converted; coroutine
  cancellation cannot be suppressed accidentally.

## Tests

`RetrofitErrorParserTest` — 8 unit tests:

| Test | Scenario |
|------|----------|
| `jsonErrorBodyPreservesStatusCodeErrorCodeAndMessage` | Full JSON body parsed correctly |
| `jsonErrorBodyWithNoMessageFallsBackToErrorCode` | JSON with `errorCode` only uses it as message |
| `plainTextErrorBodyIsUsedAsMessage` | Non-JSON body passed through verbatim |
| `emptyErrorBodyUsesFallbackWithStatusCode` | Blank body → fallback + HTTP code |
| `ioExceptionProducesNetworkError` | `IOException` with message |
| `ioExceptionWithNullMessageProducesDefaultNetworkMessage` | `IOException()` with no message |
| `unknownExceptionProducesUnknownError` | `RuntimeException` etc. |
| `cancellationExceptionIsRethrown` | `CancellationException` never swallowed |

All tests pass: `./gradlew :app:testDebugUnitTest --tests "com.machikoro.client.network.RetrofitErrorParserTest"`

## Notes for reviewers

- ViewModels (`LoginDialogViewModel`, `RegisterDialogViewModel`) are **intentionally
  untouched** — wiring `toClientError()` into them is tracked in separate sub-issues.
- The `Json` instance in `RetrofitErrorParser` is private and scoped to the file;
  it does not share state with the one in `AuthApiFactory`.
- `ServerErrorBody` is a private `@Serializable` data class — not part of the public API.


Moritz2k1 (approved):
Improves error handling, looks fairly well to me and should not mess up the game logic

PR #320, opened 2026-06-10
306 fix roll dice button UI rename label and reposition dice image

kutschera13:
#### `Dice.kt`
- Extracted `DiceSection` from `GameScreen` into its own composable
- Added rolling animation (300ms per frame, 4s total) before result is revealed
- Roll Dice button disappears while rolling or animating
- Result only appears after animation completes
- 1🎲 / 2🎲 toggle shown when Train Station is built
- "Roll Dice Again" label shown when Radio Tower is built
- Non-active player sees animation with correct dice count from `diceResult.size`

#### `GameScreen.kt`
- Replaced inline dice block with `DiceSection`
- `DiceSection` visible during `ROLL_DICE` and `RESOLVE_EFFECTS`
- Left-side dice result hidden during `ROLL_DICE` and `RESOLVE_EFFECTS`

#### `GameScreenState.kt`
- Added `requestedDiceCount: Int`
- Added `rollingStartTime: Long?`
- Added `hasRadioTower` computed property

#### `GameScreenViewModel.kt`
- `rollDice()` sets `requestedDiceCount`, `isRolling` and `rollingStartTime`
- `diceResult` and `isRolling` reset on active player change

---

### Known issues
- [ ] 2-dice animation count not always correctly shown during rolling phase – when the player selects 2🎲 and rolls, only 1 animated die is displayed instead of 2. The server correctly returns `result:[x,y]` for 2-dice rolls and `requestedDiceCount` is set correctly in the ViewModel, but `animationDiceCount` in `DiceSection` does not always reflect the correct value during the animation. Root cause not yet identified – further investigation needed.


doniliks (dismissed):
1. Dice display: display dices based on train staition status. Replace buttons with numbers to 1/2 dices. They have to be clickable and number of chosen is saved and send to server after roll dice.2. Sum without " = X", just number see sketches.
3. Show animation not while true but during eg 5 seconds.
4. Change dices faster
5. Remove is rolling - dice display has to change subphase automatically without flag for isRolling. 
6. Remove start time as well
7. Dont save local values, they are stored in state.
8.     if (state.gamePhase == GamePhase.ROLL_DICE || state.gamePhase == GamePhase.RESOLVE_EFFECTS) {
                        DiceSection(
                            state = state,
                            onRollDice = onRollDice,
                            modifier = Modifier.align(Alignment.Center)
                        )
                    } is in center part of the screen during GamePhase.RESOLVE_EFFECTS as well, this will be fixed later so keep rolliing phase only

9. Make sure that reroll is possible only once

So basically your class should call different functions for this screen:
Rolled number composable with selected diced - number from state, last chosen number from this class. Reset it after last rolled number is null.
Then call animation, for some seconds only, add delay!
Show dice after its not null.
After that we should call send resolve effect automaticly - will be done later.

Add timers above roll dice/ reroll button, you find it in shared UI. Make it 5 sec or something (add conatant at the top)  we will align them later, now for UI only with no functionality.

eles17 (dismissed):
Thanks for the work here. I don’t think this is ready to merge yet.

Main blocker: the active `GameScreen` roll-dice UI still uses the old inline implementation, so issue #306 is not actually fixed in the screen users see. The button still shows `Würfeln` / `Nochmal würfeln`, and the dice image is still passed as `leftIcon` inside the button. The new `DiceSection` exists, but it is not wired into `GameScreen`.

The PR description also does not match the current diff: it says `GameScreen.kt` replaced the inline dice block with `DiceSection`, and that `rollDice()` sets `requestedDiceCount` / `rollingStartTime`, but the current code does not do that. Those state fields are currently added but unused.

Please either:
- wire `DiceSection` into the `ROLL_DICE` branch and remove the stale inline dice UI, or
- keep this PR scoped to #306 and simply change the visible button label to `Roll Dice` and move the dice image outside the button.

Also please remove the unrelated `gradle/verification-metadata.xml` churn unless it is intentionally required.

valentinaschiavon99 (changes requested):
## Request changes

Thanks for the work on the dice UI. I compared this PR against the Server `main` branch and found a few blocking bugs, all stemming from the same root cause. Concrete fixes are listed below.

**Note on process:** this PR became hard to review because of its size (+1024 / -744, with `GameScreen.kt` almost fully rewritten and changes spread across 4 files plus `verification-metadata.xml`). It would have been better to split it into separate sub-issues, for example: (1) extracting `DiceSection`, (2) the roll animation, (3) the 1/2 dice Train Station toggle, (4) the Radio Tower reroll. Smaller, focused diffs are far easier to review and test, and would likely have surfaced the bugs below earlier.

---

## Root cause: the client never clears `diceResult` between turns

The Server is correct and authoritative. On every turn change it clears the dice state (`GameDao.advanceTurn`):

```kotlin
it[Games.lastDiceRoll] = null
it[Games.lastDiceCount] = null
```

So every start-of-turn snapshot the server sends carries `lastDiceRoll: null`.

But the client snapshot handler (`OkHttpWebSocketClient.kt:952`) only sets `diceResult` when a value is present, and never clears it:

```kotlin
game.optIntOrNull("lastDiceRoll")?.let { mutableDiceResult.value = listOf(it) }
```

The only place `mutableDiceResult` is reset to `null` is `resetGameState()` (line 1354), i.e. only on disconnect. As a result, after the first roll of the game, `diceResult` stays non-null forever. This PR introduces logic in `DiceSection` that critically depends on `diceResult == null` between turns, turning this latent bug into functional errors.

### Recommended fix (resolves most of the issues)

Make the client honor the cleared state from the server. In `OkHttpWebSocketClient.kt:952`, assign unconditionally:

```kotlin
mutableDiceResult.value = game.optIntOrNull("lastDiceRoll")?.let { listOf(it) }
```

This makes `diceResult` return to `null` at the start of each turn (since the server sends `lastDiceRoll: null`), restoring the Roll button and making the reset in `DiceSection` actually work.

---

## Bug 1: the Roll Dice button disappears on later turns (blocking)

In `DiceSection` the button only renders when:

```kotlin
if (state.gamePhase == GamePhase.ROLL_DICE && state.diceResult == null)
```

Because `diceResult` is never cleared between turns, from the active player's second turn onward this condition is `false`: the button does not appear and the stale previous result is shown instead. The player can no longer roll.

Fix: apply the root-cause fix above. Alternatively/in addition, clear `diceResult` in the `ViewModel` when the phase becomes `ROLL_DICE` or on `activePlayerId` change.

---

## Bug 2: the per-turn reset in `DiceSection` is dead code

```kotlin
LaunchedEffect(state.diceResult) {
    if (state.diceResult == null) {
        selectedDiceCount = null
        frozenDiceCount = null
        hasRerolled = false
    }
}
```

This branch almost never runs (see above), so `selectedDiceCount` / `frozenDiceCount` / `hasRerolled` are only reset by `remember(state.roundNumber)`. But in Machi Koro `roundNumber` changes once per full round (after every player has gone), not on every turn. So within the same round a player's 2-dice selection and `hasRerolled` flag persist across their turns. This is very likely the cause of the "known issue" already documented in the PR about the wrong dice count during the animation.

Fix: after applying the root-cause fix (`diceResult` returns to `null` at the start of each turn), this `LaunchedEffect` will work correctly. In addition, I recommend changing the key from `remember(state.roundNumber)` to one that changes on every turn (e.g. `roundNumber` + `activePlayerId`):

```kotlin
var selectedDiceCount by remember(state.roundNumber, state.activePlayerId) { mutableStateOf<Int?>(null) }
var frozenDiceCount by remember(state.roundNumber, state.activePlayerId) { mutableStateOf<Int?>(null) }
var hasRerolled by remember(state.roundNumber, state.activePlayerId) { mutableStateOf(false) }
```

---

## Bug 3: after reconnect the dice are wrong for 2-dice rolls

The server persists only the total (`lastDiceRoll`), not the individual dice. On sync the client builds `diceResult = listOf(total)` (a single element). The new code uses `animationDiceCount = state.diceResult?.size`, which after a reconnect equals `1` even for a 2-dice roll, and `DiceResultDisplay` shows a single face for the total value (e.g. 7 falls back to the perspective image).

Fix: this cannot be reconstructed on the client from the snapshot alone. Two options: (a) ask the Server to expose the individual dice in the snapshot; or (b) accept showing the total after reconnect and do not base `animationDiceCount` on `diceResult.size` when the value comes from sync.

---

## Other minor points

- Inaccurate PR description: it states that "`diceResult` and `isRolling` are reset on active player change", but the `ViewModel` diff only changes the `rollDice()` line. That reset was not added. This is exactly the cause of Bug 1.
- Animation timing: the animation uses a fixed 5s client-side timer (`DICE_ANIMATION_DURATION_MS`) independent of the server. If the server response arrives after 5s, there is a window where `isAnimating=false` and `diceResult` is still stale. Correctness is protected by the server guard `ROLL_ALREADY_COMPLETED`, but the UX suffers.
- Doc nit: the description says "300ms per frame, 4s total", the code uses a 100ms interval / 5s.

---

## What is correct (verified against Server `main`)

- Send envelope: the client sends `{type, gameId, payload:{gameId, diceCount}}`, and the server reads `payload.diceCount` via `RollDiceRequest.resolveDiceCount`.
- Destinations `/app/game.rollDice` and `/app/game.rerollDice` both exist on `main`.
- Result parsing: the server sends type `ROLL_DICE` with `payload.result` for both `DICE_ROLLED` and `DICE_REROLLED`; the single `parseDiceResult` handles both.
- Reroll: the client sends a count but the server ignores it and uses the stored `lastDiceCount`, so there is no conflict.
- Phase enums are aligned between client (`GamePhase`) and server (`TurnPhase`).

In short: the server's dice state is handled correctly. The client just needs to stop ignoring the cleared state from the snapshot.

valentinaschiavon99 (dismissed):
## Review — requesting changes 🔧

Nice work on the dice rework — the per-die rendering, the train-station selector, and the purple-establishment buy guard all look good and are well tested. A few things should be addressed before merge, plus one item I'd like confirmed against the server.

### 🔴 Must confirm — server timing for `lastDiceRoll`
`OkHttpWebSocketClient.applyGameStateSnapshot` now assigns the dice result unconditionally:
```kotlin
mutableDiceResult.value = game.optIntOrNull("lastDiceRoll")?.let { listOf(it) }
```
Clearing on `null` is the right fix, but this snapshot handler runs on **every** `GAME_ACTION`/`ROLL_DICE` message, not just on reconnect. It is only safe if the server keeps `lastDiceRoll` populated through `RESOLVE_EFFECTS` and nulls it exactly at `advanceTurn` (as the comment claims). If the server nulls it earlier, the dice result will disappear from the screen prematurely. Please confirm the server's actual behaviour (or link where it's nulled) before merging.

### 🔴 Animation is local-only and decoupled from server state
- `isAnimating` is local `DiceSection` state set **only** inside the roll/reroll `onClick`. Non-active players never trigger it, so the rolling animation never plays for them — which contradicts the PR description ("Non-active player sees animation with correct dice count from `diceResult.size`"). Please either implement that path or correct the description.
- The animation runs a fixed `DICE_ANIMATION_DURATION_MS = 5000` regardless of server response. The active player waits the full 5 s even when the result returns in ~200 ms, and if the server takes >5 s the `when` falls through to `else -> {}`, leaving an empty gap until the result lands. Consider driving the animation from `state.isRolling` / the server result instead of a fixed local timer — that would also fix the two items below.

### 🟠 Dead state: `rollingStartTime`
`rollingStartTime` is written in `rollDice()` and reset in `clear()` but never read anywhere (the animation uses local `isAnimating`). Please remove it, or wire it into a server-driven animation.

### 🟠 Known "2 dice shows 1" issue
This is most likely a recomposition race: `LaunchedEffect(state.diceResult)` resets `frozenDiceCount` to `null`, and the fallback `frozenDiceCount ?: selectedDiceCount ?: state.requestedDiceCount` can momentarily resolve to a lagging `requestedDiceCount` (default 1). Worth resolving (or at least tracking in an issue) rather than merging as a known bug.

### 🟡 Pre-existing latent bug worth a follow-up (non-blocking)
Because the server snapshot only stores the dice **total** (`lastDiceRoll`), the next `GAME_ACTION` snapshot overwrites the live `[x, y]` with `[total]`, so `DiceResultDisplay` collapses to a single die (falling back to the perspective image for totals 7–12). This isn't introduced by this PR, but the new per-die rendering makes it user-visible. A follow-up issue would be good.

### ✅ Looks good
- Reroll wiring (`MainActivity`/`AppRoot` → `rerollDice` / `canRerollThisTurn`).
- `isAlreadyOwnedPurpleEstablishment` logic + tests.
- Left-side `DiceResultDisplay` correctly hidden during `ROLL_DICE`/`RESOLVE_EFFECTS` (no double render).

valentinaschiavon99 (changes requested):
## Follow-up review

Thanks for the updates. The latest commits address most of the previous feedback:

- **Non-active player animation** is now implemented via `LaunchedEffect(state.diceResult)`, capturing the dice count from the live result and playing a shorter animation (`NON_ACTIVE_DICE_ANIMATION_MS`) before reveal.
- **Animation no longer blocks on a fixed timer.** `LaunchedEffect(state.isRolling)` stops the active player's animation as soon as the server confirms the result, and the new `DICE_ROLL_TIMEOUT_MS` safety net in the ViewModel prevents the animation from getting stuck.
- **`rollingStartTime`** has been removed.
- The **dice-count race** is mitigated now that `frozenDiceCount` is set on the roll action and takes priority in `animationDiceCount`.

The ViewModel rework is solid — guarding against concurrent taps (`if (isRolling) return`), the roll timeout, and the phase-change cancellation (#175) all look correct.

Two items remain before I can approve:

### 1. Confirm server timing for `lastDiceRoll` (blocking)
This line in `OkHttpWebSocketClient.applyGameStateSnapshot` is unchanged from the previous review:
```kotlin
mutableDiceResult.value = game.optIntOrNull("lastDiceRoll")?.let { listOf(it) }
```
The handler runs on every `GAME_ACTION`/`ROLL_DICE` snapshot, not only on reconnect. Clearing on `null` is correct only if the server keeps `lastDiceRoll` populated through `RESOLVE_EFFECTS` and nulls it at `advanceTurn`. If it is nulled earlier, the result will disappear from the screen prematurely. Please confirm the server behaviour (or link where `lastDiceRoll` is cleared) so we can be sure this is safe.

### 2. Non-active animation re-triggers on result overwrite (should fix)
The server snapshot stores only the dice total, so a follow-up `GAME_ACTION` snapshot overwrites the live `diceResult` (e.g. `[3, 4]` becomes `[7]`). For a non-active player this changes `state.diceResult`, which re-fires `LaunchedEffect(state.diceResult)` and replays the animation a second time (with a single die) before re-revealing the result. Please guard the non-active animation so it only plays on a genuinely new roll rather than on every `diceResult` change.

valentinaschiavon99 (approved):
## Approved

The code-side feedback from the previous reviews has been fully addressed:

- Non-active player animation now plays once per turn, guarded by `hasAnimatedThisTurn`, so the snapshot overwrite of the live result (`[x, y]` becoming `[total]`) no longer replays it.
- The animation is driven by `state.isRolling` and stops as soon as the server confirms the result, with `DICE_ROLL_TIMEOUT_MS` as a safety net.
- The dead `rollingStartTime` field was removed, and the dice-count race is mitigated via `frozenDiceCount`.
- The ViewModel guards against concurrent rolls, handles the roll timeout, and cancels rolls on phase change (#175).

Approving on that basis. Two items to keep in mind, neither blocking:

1. The `lastDiceRoll` handling in `OkHttpWebSocketClient.applyGameStateSnapshot` relies on the server keeping the value populated through `RESOLVE_EFFECTS` and clearing it at `advanceTurn`. This is tracked separately in the thread ("wait until the server is adjusted"); please confirm the server behaviour before relying on it in production.
2. `hasAnimatedThisTurn` is not reset in the `diceResult == null` branch, so a non-active player will not see a Radio Tower reroll animate within the same turn. The result still displays correctly, so this is a minor follow-up.

PR #321, opened 2026-06-10
fix: show live marketplace quantities in buy shop

eles17:
## Summary
- Show live remaining marketplace quantities on buy/build establishment shop cards.
- Use marketplace state as the source of truth for establishment availability in shop selection and purchase guards.
- Visually mark zero-stock establishment cards as unavailable.

## Testing
- `./gradlew test`


doniliks (dismissed):
Please reuse card render and quantity indicator from own player cards row/grid to keep UI consistent. If possible, extract existing quantity indicator to own class for easier UI changes. Also make sure that top padding is adjusted same way to keep current cards size.

valentinaschiavon99 (dismissed):
The client↔server contract checks out — the marketplace map shape (keyed by CardType enum names), its location at payload.state.marketplace, the live re-broadcast after each purchase, and the sold-out-as-0 behavior all match the server's current main. Nice work on the live-quantity wiring.

One change needed before merge:

Shop.kt — operator-precedence bug in contentDescription (lines 179-182)

contentDescription = "...${item.effectText}" +
        remainingQuantity?.let { ", $it remaining" }.orEmpty() +
        if (remainingQuantity == 0) ", unavailable" else "" +
        if (isRecommended) ", recommended" else ""
Kotlin's else is greedy, so this parses as:

... + if (remainingQuantity == 0) ", unavailable"
     else ("" + if (isRecommended) ", recommended" else "")
The ", recommended" suffix ends up nested inside the else branch of the first if, so it's silently dropped whenever a card is sold out (remainingQuantity == 0). Please wrap each conditional in parentheses so the three suffixes append independently:

contentDescription = "...${item.effectText}" +
        remainingQuantity?.let { ", $it remaining" }.orEmpty() +
        (if (remainingQuantity == 0) ", unavailable" else "") +
        (if (isRecommended) ", recommended" else "")
Everything else (ViewModel purchase gates, empty-marketplace fallback to item.isAvailable, the CardArtImage/CardQuantityIndicator extraction) looks good.

valentinaschiavon99 (approved):
Approved ✅

Verified the client↔server contract against the server's current main — everything lines up:

CardType enum — identical 15 values/names on both sides, so CardType.valueOf(item.type) round-trips cleanly.
Marketplace shape — server's GameStateDto.marketplace: Map<CardType, Int> serializes with enum-name keys ("BAKERY", etc.), exactly what parseMarketplace/remainingMarketplaceQuantityFor expect.
Envelope — quantities are read from payload.state.marketplace, matching buildGameActionPayload's "state" snapshot.
Live updates — purchase() decrements supply then broadcastPurchase() re-broadcasts a fresh snapshot via GAME_ACTION, so the shop reflects quantities in real time.
Sold-out handling — the server floors quantity at 0 and never drops the row, so marketplace[cardType] ?: 0 correctly renders a depleted card as unavailable. Landmarks return null and fall back to item.isAvailable, so no landmark regression.
The contentDescription operator-precedence issue from the earlier revision is fixed in 435fe447 (both conditionals now parenthesized), so the ", recommended" suffix is no longer dropped on sold-out cards.

The ViewModel purchase gates switched from item.isAvailable to isShopItemAvailableFromMarketplace(item) consistently, the empty-marketplace fallback is sound, and the CardArtImage/CardQuantityIndicator extraction is behavior-preserving with good test coverage. Nice work. 🎉

PR #323, opened 2026-06-11
feat(game): auto-advance the Resolving Effects phase (#302)

FabianSolutions:
## Summary

Implements **#302** — the client's `RESOLVE_EFFECTS` turn phase now **auto-advances**. After a dice roll the client dwells on Resolving Effects for a configurable **~20s** and then continues to the buying phase on its own, instead of the active player having to press a manual "Resolve effects" button.

Part of the **#301** resolving-effects epic. The rich effect *view* (#303), coin *animations* (#304), and the server effect-detail *payload* (#305) are intentionally **out of scope** here — this PR is the turn-flow plumbing only.

## Changes

- **`GameScreenViewModel`** — a state-driven observer auto-sends `resolveEffects(gameId)` once the local player is the active player in `RESOLVE_EFFECTS`, after `DEFAULT_RESOLVE_EFFECTS_DWELL_MS` (20s; injectable via constructor for tests/tuning). It's driven off the aggregated `state` (immune to the order in which phase/activePlayer/status/gameId settle), `distinctUntilChanged` (arms exactly once per entry), with a post-delay re-check and cancel-on-leave so a stale timer never fires. Only the **active player** sends; everyone else waits for the server's broadcast, so all clients dwell on the phase together. The `RESOLVE_EFFECTS` branch is dropped from `performTurnFlowAction`.
- **`GameScreen`** — the manual "Resolve effects" button is gone (`turnFlowActionLabel` returns null for that phase). The existing phase banner still shows "Resolve Effects".

**No server change.** Reuses the existing `/app/game.resolveEffects` contract (the same call the manual button made); the server stays authoritative for the actual switch to `BUY_OR_BUILD`. Verified the server blocks in `RESOLVE_EFFECTS` waiting for that frame — no force-advance timer — so the client dwell is safe.

## Acceptance criteria — all covered

- ✅ After dice are rolled, the client shows a resolving-effects phase (existing phase banner).
- ✅ The old manual "Resolve effects" button is removed.
- ✅ The client automatically proceeds to buying after the dwell.
- ✅ Delay is configurable, ~20s default.
- ✅ Stays consistent with the server's turn/phase model (no contract change).

## Testing

- ✅ `:app:assembleDebug` + `:app:testDebugUnitTest` green — **73/73** in `GameScreenViewModelTest`, incl. 7 new cases: auto-sends after the dwell; not before (`advanceTimeBy` boundary); not for a non-active player; not when not `IN_PROGRESS`; cancelled if the phase leaves mid-dwell; fires once per entry; manual `performTurnFlowAction` no longer resolves.
- ✅ `:app:lintDebug` — 0 errors.
- ✅ Adversarial code review (double-send, stuck, `distinctUntilChanged` turn boundaries, job scoping/leak, post-purchase-endTurn conflict, fire-while-inactive/after-end) — no issues.
- ✅ **Live 2-emulator demo**: active player rolled → entered Resolving Effects with **no manual button** → after ~20s with zero input it **auto-advanced to Buy or Build** (DB- and screenshot-confirmed).

## Notes

The dwell is purely time-driven for now, so the client waits the full ~20s even when nothing actually triggered — this matches the issue's "constants now, visual timers later" guidance and will be refined when #303/#304 land.

Refs #302


Moritz2k1 (approved):
Auto advances the resolve effects phase now. LGTM. Time has to tweaked, 20s is a bit long

PR #334, opened 2026-06-13
fix: require numeric activation numbers for card definitions

eles17:
## Summary
This finalizes #307 under the parent epic #300 by making activation numbers numeric-only for structured client logic.

The client already models shop/card activation values as `List<Int>` and derives `activationText` for UI display. This change tightens the websocket parsing boundary so server-provided card definitions only populate structured activation values from the numeric `activationNumbers` field. Legacy/display-oriented fields such as `activationText` or `activationRange` are no longer parsed back into activation numbers.

This keeps the contract clean:
- `activationNumbers` is the source of truth for sorting, recommendations, and effect-related logic.
- `activationText` remains UI-only and is derived from `activationNumbers`.
- Existing shop/card display behavior stays the same.
- The marketplace/shop behavior from #321 is unchanged.

## Changes
- Removed legacy string parsing from `ShopItem.kt`.
- Updated websocket card-definition parsing to ignore `activationText` / `activationRange` as structured activation sources.
- Added parser coverage to ensure display strings are not treated as activation numbers.
- Updated card-definition tests to focus on deriving display text from numeric activation values.

### DIFF
--- a/app/src/main/java/com/machikoro/client/domain/model/shop/ShopItem.kt
+++ b/app/src/main/java/com/machikoro/client/domain/model/shop/ShopItem.kt
@@ -300,15 +300,3 @@ fun List<Int>.toActivationText(): String? {
         "${numbers.first()}-${numbers.last()}"
     }
 }
-
-fun String.toActivationNumbers(): List<Int> {
-    val trimmed = trim()
-    if (trimmed.isBlank() || trimmed.equals("Permanent", ignoreCase = true)) return emptyList()
-    val rangeParts = trimmed.split("-", limit = 2).map { it.trim().toIntOrNull() }
-    if (rangeParts.size == 2 && rangeParts.all { it != null }) {
-        val start = rangeParts[0] ?: return emptyList()
-        val end = rangeParts[1] ?: return emptyList()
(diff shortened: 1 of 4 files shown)


valentinaschiavon99 (approved):
**Approve**

Verified locally: `CardDefinitionsTest` + `OkHttpWebSocketClientTest` pass. The change correctly drops the fragile display-string parsing (`activationRange` / `activationText` → numbers) and trusts only the structured numeric `activationNumbers` array, with the canonical `CardDefinitions` set as a safe fallback. This aligns the client to the server's actual contract — every SYNC frame in the test suite sends `activationNumbers` as a numeric array — and removes a real source of client/server drift. Dead code (`String.toActivationNumbers` + import) is cleaned up.

One contract assumption to confirm with the backend, not a blocker: the client now relies on the server **always populating the numeric `activationNumbers` array for every card definition**. If a future custom/variant card ships with only `activationText`/`activationRange` and no numeric array, the client will fall back to the standard canonical values and silently show the wrong activation numbers. Worth a quick confirmation that `activationNumbers` is guaranteed for all card definitions going forward.

PR #340, opened 2026-06-13
docs(rules): document the Insider Trading cheat in the in-app tutorial (#336)

FabianSolutions:
## What & why

Closes #336. The Insider Trading cheat and the accusation mechanic are fully implemented but documented **nowhere a player can find them** — the only tutorial is the bundled `rules.pdf`, which has zero mention of the cheat. Sprint 3 requires the cheat to be in the tutorial.

This adds the documentation as **native Compose text appended as the final page of the existing in-app rules viewer** (issue Option A), so a player paging through the tutorial reaches it naturally. Option B (regenerating the 8.4 MB binary PDF) was rejected — not diff-friendly, not testable.

## Changes

- **`CheatRulesContent.kt`** (new) — a scrollable "Cheating & Accusations" page: an *Insider Trading* block (shake / **Insider tip** button → green highlight; valid one turn; reported to the server) and an *Accusing a cheater* block (**Accuse of cheating** via player inspection; **once per turn**; **wrong guess costs a coin**). Wording mirrors the live in-game labels so the docs match the UI.
- **`PdfViewerScreen.kt`** — tracks the real PDF page count separately and shows `totalPages = pageCount + 1`, rendering `CheatRulesContent` on the synthetic last page. Existing Previous/Next nav and the "Page X of Y" indicator are reused; a `pdfPageCount > 0` guard avoids a pre-load flash.
- **`CheatRulesContentTest.kt`** (new) — instrumented Compose test asserting the section heading and every acceptance-criterion detail is shown.

## Acceptance criteria

- [x] A player can discover the cheat + accusation rules from the in-app tutorial
- [x] Covers activation (shake / *Insider tip*), one-turn validity, server reporting, accusation mechanic + once-per-turn + wrong-accusation penalty
- [x] Reachable from the Home → Rules entry point (it's the last page)
- [x] Covered by a UI test (Option A)

## Verification

- `:app:assembleDebug` + `:app:lintDebug` — green.
- **Live emulator run**: registered + logged in, opened Home → Rules → paged to the end; the viewer reports **"Page 9 of 9"** and renders the cheat page, with "Next" correctly disabled on the last page.
- `CheatRulesContentTest` follows the existing `androidTest` convention (runs on a connected emulator, not in the lint/sonar CI).

### DIFF
--- a/app/src/androidTest/java/com/machikoro/client/ui/home/CheatRulesContentTest.kt
+++ b/app/src/androidTest/java/com/machikoro/client/ui/home/CheatRulesContentTest.kt
@@ -1,0 +1,57 @@
+package com.machikoro.client.ui.home
+
+import androidx.compose.ui.test.assertIsDisplayed
+import androidx.compose.ui.test.junit4.createComposeRule
+import androidx.compose.ui.test.onNodeWithText
+import com.machikoro.client.ui.theme.ClientTheme
+import org.junit.Rule
+import org.junit.Test
+
+/**
+ * Verifies the in-app cheat/accusation tutorial section (issue #336). The
+ * section is rendered as the final page of the rules viewer, so showing it here
+ * is what a player reaches from Home -> Rules -> last page.
+ */
+class CheatRulesContentTest {
+    @get:Rule
+    val composeTestRule = createComposeRule()
+
+    private fun setContent() {
+        composeTestRule.setContent {
+            ClientTheme {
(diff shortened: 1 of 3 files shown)


eles17 (approved):
Thanks for the implementation. Looks good to me. Later on we can move the rules somewhere else if wanted but i think in rules fits just well. 👍  Approved.

PR #341, opened 2026-06-14
Refactor UI 

Ivona-Goranova:
## Changes

* Replaced the player coin icon with an updated version
* Updated landmark card images with higher-quality assets
* Added image-based card selection and recommendation frames
* Added new frame assets:

  * `card_frame.png`
  * `card_frame_green.png`
* Refactored shop card highlighting to use frame images instead of simple borders
* Updated the logout/leave button styling


eles17 (dismissed):
Thanks for the UI refresh. Build, lint, and unit tests are passing, but I think we should fix one UI regression before merging.

After a purchase succeeds, `GameScreenViewModel` clears `selectedPurchaseItemType` and keeps only `purchaseFeedbackItemType`. The new shop frame rendering only checks `isSelected` / `isRecommended`, so the purchased card no longer shows the success feedback frame after server confirmation.

Please restore visual feedback for `purchaseFeedbackItemType`, ideally using the new frame asset so the success state remains visible after selection is cleared.

Suggested fix area:
`app/src/main/java/com/machikoro/client/ui/game/ui/Shop.kt`

doniliks (dismissed):
You have mixed up screens and now log out button is in lobby screen 
Also you have removed offset from coin field, please align dice as requested so coin and dice have same image and font size and centered on right left

eles17 (commented):
LGTM. @doniliks approve?!

eles17 (approved):
Approved

PR #342, opened 2026-06-14
175 fix prevent unintended screen skips and endless dice animation during gameplay

eles17:
## Summary

Fixes client-side safeguards for #175 to prevent unintended screen skips and endless dice animation during gameplay.

The client now avoids navigating from stale/hydrated state alone, requires explicit user flow before entering lobby/game screens, and ensures dice rolling cannot stay stuck forever if the server misses or delays a dice result.

## Changes

- Added explicit login gating in `NavigationViewModel`
  - Session hydration alone no longer skips the start screen.
  - Game navigation still requires valid lobby/game flow.
  - Reconnect snapshots cannot directly jump into the game screen.

- Updated `MainActivity`
  - Removed stale `activeGameId` navigation behavior.
  - Wires login/logout actions to set/reset navigation state.
  - Keeps lobby navigation tied to explicit lobby events.

- Stabilized dice rolling in `GameScreenViewModel`
  - Stops rolling when a dice result arrives.
  - Stops rolling when phase leaves `ROLL_DICE`.
  - Adds a 10s fallback timeout if the server never responds.
  - Ignores repeated roll clicks while already rolling.

- Improved lobby creation guard in `HomeViewModel`
  - Prevents duplicate create-lobby requests.
  - Releases the in-flight guard after timeout if no lobby code is received.

- Cleaned accidental branch artifacts
  - Restored default/commented backend URL config.
  - Removed accidental `tyle=official` file.

## Verification

- `./gradlew testDebugUnitTest`
- `./gradlew lint`
- `./gradlew jacocoTestCoverageVerification`

## Related Issues

Closes https://github.com/SE2-Machi-Koro/Client/issues/175

Server alignment follow-up:
https://github.com/SE2-Machi-Koro/Server/issues/378

The client now defensively handles stale or missing server state, but server #378 should align backend state/realtime events so stale `gameId`, reconnect membership, duplicate lobby actions, and round-specific dice completion issues are handled at the source.


valentinaschiavon99 (changes requested):
Solid, well-scoped fix. The navigation gating (`hasBeenInLobby` + `userHasLoggedIn`) is the right approach and is the strongest part of the change — session hydration no longer skips the start screen, and reconnect snapshots no longer jump straight into the game. Test coverage for the new branches is good. It also lines up cleanly with [Server #378](https://github.com/SE2-Machi-Koro/Server/issues/378) as the client-side defensive layer. A few things to address before merge.

## 🔴 Blocking: stale dice-roll timeout can cut off a *later* roll

`GameScreenViewModel.rollDice()` launches a fire-and-forget 10s timeout that only checks the global `isRolling` flag — it isn't bound to the roll that started it:

```kotlin
viewModelScope.launch {
    delay(DICE_ROLL_TIMEOUT_MS)
    mutableState.update { current ->
        if (current.isRolling) current.copy(isRolling = false) else current
    }
}
```

- Roll A at t=0 → launches timeout-A (fires t=10s)
- Server replies t=2s → `isRolling=false` ✅
- Roll B at t=8s (next turn) → `isRolling=true`
- t=10s: **timeout-A fires**, sees `isRolling==true` (from B), and clears B's animation while the server is still working.

Any two rolls <10s apart hit this — common in a fast game. Since this is exactly the round-N dice-completion fallback Server #378 (item 6) expects the client to provide, the bug undermines the feature's purpose. Please store the timeout in a `Job` and cancel/replace it on each new roll and on `diceResult`, or tag each roll with an id and only clear if it still matches.

`rerollDice()` is collateral: it sets `isRolling=true` with no timeout of its own, so a stale roll timeout can clear an in-flight reroll, and a lost reroll result hangs forever. Pre-existing, but worth covering with the same Job mechanism.

## 🟡 Join-tap dedupe gap (asymmetry with create)

The new dedupe guard only protects `createLobby`. `joinLobby()`'s already-connected fast path sends immediately with no in-flight guard:

```kotlin
if (webSocketClient.connectionStatus.value == ConnectionStatus.CONNECTED) {
    webSocketClient.sendJoinLobby(code)   // rapid taps → multiple lobby.join frames
    return
}
```

This leans entirely on Server #378 (item 5: idempotent join). Mirroring the create-lobby guard here would make the client self-consistent.

## 🟡 Optimistic `showLobby()` before server confirmation

`onCreateLobbyClick` and `onJoinLobbySubmit` now call `navigationViewModel.showLobby()` synchronously, so navigation no longer waits for the server's `lobbyEntered`:

- Invalid join code → flash to an empty Lobby, then `lobbyJoinErrors` → `leaveLobby()` bounces back Home. Recoverable, but a visible flash.
- Create that times out (the case this PR guards) → `isCreatingLobby` resets after 10s, but nothing navigates off the empty Lobby; there's no create-error recovery path the way join has. User can be stranded.

Consider keeping lobby navigation gated on the server event, or adding a create-failure recovery.

## Notes (non-blocking)

- **Logout comment wording** (`MainActivity`): the stale `gameId` actually travels in the `/app/chat.addUser` JOIN (`sendJoinMessage`), not the `lobby.create` JOIN. The fix itself is correct (`clearLobbyCode()` nulls `activeGameId` via `resetLobbyState()`), just tighten the comment so the next reader looks in the right place. Note a *plain reconnect* still resends `activeGameId`, so the client remains dependent on Server #378 item 1 there — fine as a defensive split, worth a one-line acknowledgment.
- **Process-death restore**: `userHasLoggedIn` lives only in the ViewModel while the session persists across process recreation, so a killed-then-restored user with a valid token lands on Start and must log in again. This looks like the intended trade-off — just confirm it's acceptable.

## Verdict

Requesting changes on the dice-timeout bug; the rest are quick follow-ups or judgment calls. Nice work on the navigation gating.

valentinaschiavon99 (approved):
All previously raised points have been addressed, and the full set of affected unit tests passes locally (135 tests, 0 failures across GameScreenViewModelTest, HomeScreenViewModelTest, and NavigationViewModelTest).

Summary of resolutions:

- Dice roll timeout: reworked with a per-roll identifier and expected-phase tracking. The timeout now clears `isRolling` only when it still owns the active roll, and is explicitly cancelled on `diceResult` and on the phase-change failsafe. This resolves the case where a stale timeout from a completed roll could cancel a subsequent roll's animation.
- Reroll: now guarded against re-entry while a roll is in progress and covered by the same timeout mechanism, closing the previous gap where a lost reroll result could leave the animation running indefinitely.
- Lobby join: added an in-flight guard symmetric with create, with release on success, server rejection, and timeout, preventing duplicate `lobby.join` frames from rapid taps.
- Navigation: removed the optimistic `showLobby()` calls on create/join. Navigation is now driven solely by the server `lobbyEntered` event. Create still navigates correctly because the host auto-joins and the resulting `LOBBY_JOINED` emits the event.
- Logout comment: corrected to reference the `/app/chat.addUser` reconnect path and Server #378 ownership.

Test coverage added for the regressions is appropriate, including a case that reproduces the original stale-timeout scenario.

One minor, non-blocking note for a future cleanup: in the `gamePhase` collector, the `shouldCancelPendingRoll` flag is mutated inside the `MutableStateFlow.update { }` lambda. Since `update` may re-invoke its lambda under contention, it is preferable to avoid side effects there and derive the decision from the update result. It is harmless as written because the flag is only ever set to true and the cancellation is idempotent.

Approving. Nice work on the fixes and the added tests.

PR #345, opened 2026-06-15
fix: now displays all players in row, additionally winner and rounds …

doniliks:
…number played
now works players are displayed animation etc all staff

but navigation doesn't work properly now cause view leaderboard sends player to lobby @Moritz2k1 if you will adress navigation in further issues please have a look


eles17 (commented):
Thanks for the fix. I’d still request one fix before merging:

`GameOverOneWinner` now returns immediately when `rankedPlayers` is empty. That can make the winner route render a completely blank screen: no background, no header, no buttons. It also conflicts with the existing `AppRootTest.showsWinnerScreenWhenGameStatusIsFinished`, which expects the winner screen header to be visible for a finished game even with minimal state.

Please render a fallback winner/end screen instead of returning early, or make sure `AppRoot` only navigates to the winner route once the ranked player data is available.

Small cleanup: `resolveWinnerName` is still imported in `AppRoot.kt` but no longer used.

eles17 (approved):
Looks good to me now. The previous blank winner-screen issue is fixed, the unused import cleanup is done, and the PR is mergeable.

PR #351, opened 2026-06-15
325 create composables for resolve effects phase

doniliks:
Added stacks with cards including coin to represent earned amount, has to be adjusted in #303


eles17 (dismissed):
Thanks, this is close. I’d hold off merging until one small Compose issue is fixed and CI finishes.

`EffectsResolutionColumn` accepts a `modifier`, but it does not apply it to the root `Column`. Instead, the same modifier is reused on both `CoinBadge` and `LazyColumn`, which means caller-provided padding/size/test tags/background/etc. would be applied to multiple children rather than the composable as a whole.

Suggested fix: apply `modifier = modifier` to the root `Column`, and use fresh child modifiers for the badge and list.

eles17 (commented):
I think is mergable @Ivona-Goranova pls check

Ivona-Goranova (approved):
Approved. The modifier handling is fixed now: the root `Column` receives the passed-in `modifier`, and the child composables use fresh `Modifier` instances.

PR #352, opened 2026-06-15
328 add chat message model and websocket flow

SoupCube:
### Summery 
Added a ChatMessage Model 
Added Websocket Flow - send and receive chat messages
Added the parsing for chat messages
Added and adjusted Tests to fit the changes and additions


eles17 (dismissed):
Thanks for the chat websocket work. CI is green and the scope looks focused, but I don’t think this is ready to merge yet.

`sendChatMessage()` sends the chat text as `content`, while `handleChatMessage()` only reads `message` from the top level or payload. If the backend broadcasts the same DTO shape the client sends, incoming chat messages will be silently dropped.

Please either:
- send the field as `message`, if that is the agreed server contract, or
- make `handleChatMessage()` accept `content` as well, preferably with a test for that payload shape.

valentinaschiavon99 (dismissed):
Thanks for wiring up the chat flow! The receive-side parsing is in good shape, but as written this PR talks to the **old global** chat endpoint, not the new per-game endpoint added in Server #381 (https://github.com/SE2-Machi-Koro/Server/pull/381). With both PRs merged as-is, messages broadcast globally to `/topic/public` with no participant check, and Server #381's `/app/game.chat.send` is never called. Three changes are needed to align the contracts:

### 1. Send to the per-game destination, not the global one
`sendChatMessage()` currently posts to `WebSocketContract.chatSendDestination` (`/app/chat.send` → old global handler → `/topic/public`). It needs to target the new mapping `@MessageMapping("/game.chat.send")`:

```kotlin
// WebSocketContract.kt — add:
const val gameChatSendDestination: String = "/app/game.chat.send"
```
…and use it in the `StompFrame` `destination` header.

### 2. Match the server payload field — `message`, not `content`
Server expects `GameChatRequest { gameId: Int, message: String }`. `message` is non-null with no default, so a `content` field deserializes to a missing-parameter error and nothing broadcasts. Fix the send body:

```kotlin
val body = JSONObject()
    .put("gameId", gameId)
    .put("message", message)   // was: .put("content", message) + type/sender
    .toString()
```
Drop the client-supplied `type`/`sender` — the server authenticates the sender from the STOMP principal and ignores client-provided identity.

### 3. Filter the receive handler on `type == "CHAT"`
`handleChatMessage(json)` runs on every inbound frame and emits for any message carrying `sender` + `content`. Since `/topic/game/{gameId}` and `/topic/public` carry JOIN/LEAVE and other broadcasts with those fields, non-chat events will leak into the chat stream. Add a guard:

```kotlin
private fun handleChatMessage(json: JSONObject) {
    if (json.optString("type") != "CHAT") return
    // …existing parsing…
}
```
The existing tests pass only because they happen to send `type:"CHAT"`. Please add a negative case (e.g. a `JOIN` frame with content) that asserts no emission.

### Also confirm the subscription
The server broadcasts to `/topic/game/{gameId}`, which `OkHttpWebSocketClient` already subscribes to via `subscribeToGameTopic`, so received messages should land correctly once the send path is fixed. Please verify in-game chat round-trips end-to-end after these changes.

Once the destination + `message` field + `type` filter are in, the two PRs will be fully aligned. 👍

valentinaschiavon99 (approved):
Changes resolved: now sends to /app/game.chat.send with the matching `message` field, and the receive handler filters on type=="CHAT". Server #381 is merged, so the target endpoint exists. Aligned end-to-end — approving.

PR #354, opened 2026-06-17
350 fix move radio tower reroll decision before resolve effects

eles17:
## Summary

This PR fixes the Radio Tower reroll flow so the reroll decision is presented with the dice result instead of as part of the resolving-effects UI.

The backend/server phase still remains `RESOLVE_EFFECTS`, but the client now treats the Radio Tower choice as a dice-result decision before effect resolution continues.

## Changes

- Moves the Radio Tower reroll controls into the central dice roll/result area.
- Hides the side dice-result display while the Radio Tower decision UI is active, avoiding duplicate dice displays.
- Keeps the Train Station dice-count selector limited to the initial `ROLL_DICE` phase.
- Preserves the original dice count when rerolling by using the current dice-result size.
- Adds a `Skip` action so the active player can accept the current dice result.
- Wires `onSkipReroll` through `MainActivity`, `AppRoot`, and `GameScreen`.
- Prevents automatic `resolveEffects` from firing while the active player still has a pending Radio Tower reroll decision.
- Sends `resolveEffects` immediately when the player skips the reroll.
- Keeps the existing server-authoritative reroll endpoint and once-per-turn guard.

## Acceptance Criteria Covered

- Radio Tower reroll button is no longer displayed as a separate resolving-effects control.
- Reroll button appears next to the dice result.
- Players without Radio Tower continue through resolving effects normally.
- Players with Radio Tower can make one reroll decision per turn.
- Resolving-effects auto-advance waits until the player rerolls or skips.
- Resolving-effects UI remains focused on resolving effects, not dice reroll controls.


doniliks (approved):
lgtm

PR #355, opened 2026-06-17
Fix Roll Dice button UI: rename label and reposition dice image (#306)

valentinaschiavon99:
## Overview

Reworks the dice‑roll UI on the game screen and incorporates several rounds of review feedback. Closes:

- **#306** — Roll Dice button UI: rename label and reposition the dice image.
- **#346** — Non‑active players do not see the Radio Tower reroll animation within the same turn.

The branch starts from the revert of #320 (reverted in #348), **merges current `main`**, then re‑applies the dice work and the review fixes on top.

---

## #306 — Roll Dice button label & dice image
- Label reads **"Roll Dice"** (and **"Roll Dice Again"** for a reroll) instead of the German "Würfeln".
- Per review, the standalone die image previously shown beside the button was **removed** — the button stands on its own.

## #346 — Reroll animation for non‑active players
Non‑active players now animate on **every genuine roll and same‑turn reroll**, driven by a monotonic `diceRollTick` that is bumped only on real `DICE_ROLLED`/`DICE_REROLLED` frames (never on a reconnect snapshot). This animates real rerolls while ignoring the `[x, y] → [total]` snapshot collapse, order‑independently.

- `WebSocketClient` exposes `diceRollTick`; `OkHttpWebSocketClient` bumps it only on the accept path of a genuine roll/reroll frame.
- `GameScreenViewModel` folds it into `GameScreenState.diceRollTick`.
- `DiceSection` animates non‑active players when the tick advances (seeded with the current tick so entering mid‑turn doesn't replay an old roll).

---

## Review feedback addressed

**Roll dice timing / phase delay**
- Unified the rolling animation to a single brief duration (~1.5s) for active and non‑active players, so reveals stay in lockstep — which matters during a Radio Tower reroll where both spin off the same roll tick. The active player's spin still ends the instant the server confirms the result.
- The RESOLVE_EFFECTS auto‑advance dwell dropped from **20s → 6s** and now **drives the phase timer's countdown** (`RESOLVE_EFFECTS_TIMER_SECONDS`), so the number players see is the real time until the phase ends. Players without a Radio Tower no longer wait 20s with nothing to decide.

**Missing Roll Dice button / stale dice on screen**
- The server never sends a `null` `diceResult` between turns, so the previous turn's result lingered into the next `ROLL_DICE` phase — hiding the Roll Dice button (its condition is `diceResult == null`) and leaving a frozen die + number on screen. The ViewModel now clears `diceResult` when a new `ROLL_DICE` phase begins.

**Inconsistent dice display (one die vs two)**
- A live roll sets `diceResult = [x, y]`, but a routine snapshot only stores `lastDiceRoll` (the total) and was overwriting it with `[x+y]` — a single generic die — mid‑turn. The snapshot now yields to a per‑die result that already sums to that total, keeping the two‑dice display stable.

**Doubles not visible**
- `DiceResultDisplay` now shows a **"Doubles!"** label when two equal dice are rolled (e.g. relevant for the Amusement Park's extra turn).

---

## Train Station / reroll controls (unchanged behaviour, kept)
- Train‑Station 1/2 dice‑count selector during `ROLL_DICE`.
- Radio Tower reroll (#326): **"Roll Dice Again"** + **"Skip"** during `RESOLVE_EFFECTS`.

## Changed files
- `ui/game/ui/Dice.kt` — `DiceSection`, `DiceCountSelector`, animated `DiceAnimationDisplay`, per‑die `DiceResultDisplay` (+ doubles label); removed the die image beside the button; unified animation duration.
- `network/websocket/WebSocketClient.kt` + `OkHttpWebSocketClient.kt` — `diceRollTick`; snapshot no longer collapses a per‑die result into the total.
- `domain/model/state/GameScreenState.kt` — `requestedDiceCount`, `diceRollTick`.
- `ui/game/GameScreenViewModel.kt` — `rollDice()` sets `requestedDiceCount`; collect `diceRollTick`; clear stale `diceResult` on entering `ROLL_DICE`; shorter resolve dwell wired to the phase timer.
- `ui/game/GameScreen.kt` — render `DiceSection` for `ROLL_DICE`/`RESOLVE_EFFECTS`; phase timer reads the resolve dwell.
- Tests — `diceRollTickIncrementsInStateOnEachRoll`, `diceResultIsClearedWhenEnteringRollDicePhase`, `laterSnapshotDoesNotCollapsePerDiceResultIntoTotal`, plus test‑double updates.

## Verification
- `:app:compileDebugKotlin`, `:app:compileDebugUnitTestKotlin` compile.
- `:app:testDebugUnitTest` — **603 tests, 0 failures**.

Closes #306
Closes #346

### DIFF
--- a/app/src/main/java/com/machikoro/client/domain/model/state/GameScreenState.kt
+++ b/app/src/main/java/com/machikoro/client/domain/model/state/GameScreenState.kt
@@ -29,6 +29,10 @@ data class GameScreenState(
     // Short feedback text shown in the shop for pending, success, or retryable errors.
     val purchaseMessage: String? = null,
     val isRolling: Boolean = false,
+    // Dice count the active player requested for the in-flight roll. Drives the
+    // rolling animation (how many dice spin) since the snapshot persists only
+    // the total, making diceResult.size unreliable after a reconnect.
+    val requestedDiceCount: Int = 1,
     // Reconnect snapshot fields (from /app/game.sync).
     val gameStatus: GameStatus? = null,
     val roundNumber: Int? = null,
@@ -88,6 +92,7 @@ data class GameScreenState(
             purchaseFeedbackItemType = null,
             purchaseMessage = null,
             isRolling = false,
+            requestedDiceCount = 1,
             gameStatus = null,
             roundNumber = null,
             playerCards = emptyMap(),
--- a/app/src/main/java/com/machikoro/client/domain/model/state/GameScreenState.kt
+++ b/app/src/main/java/com/machikoro/client/domain/model/state/GameScreenState.kt
@@ -16,6 +16,10 @@ data class GameScreenState(
     val customDisplayText: String? = null,
     val players: List<PlayerCoinState>,
     val diceResult: List<Int>? = null,
+    // Bumped on every genuine roll/reroll (not on reconnect snapshots) so the UI
+    // can animate non-active players for each real roll, including a same-turn
+    // Radio Tower reroll (#346).
+    val diceRollTick: Long = 0L,
     val activePlayerId: Int? = null,
     val winnerId: Int? = null,
     val myUserId: Int? = null,
--- a/app/src/main/java/com/machikoro/client/network/websocket/OkHttpWebSocketClient.kt
+++ b/app/src/main/java/com/machikoro/client/network/websocket/OkHttpWebSocketClient.kt
@@ -1015,9 +1015,17 @@ class OkHttpWebSocketClient(
         parseTurnPhase(game.optString("turnPhase"))?.let { mutableGamePhase.value = it }
         game.optIntOrNull("roundNumber")?.let { mutableRoundNumber.value = it }
         // The snapshot persists only the dice total (lastDiceRoll), not the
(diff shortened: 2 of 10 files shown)


doniliks (dismissed):
1. Button roll dice has a dice in the left, remove it.
2. Timings for dice roll phase are not aligned with current timer and it takes to long till players can continue with next phase.
3. Somethings there is no button for roll dice only stalled dices and last rolled number is still displayed on the screen. 
4. Dice animation time is different for active/non active player which may cause problems is animation is active but player has clicked on reroll

doniliks (dismissed):
Delay is still extremely long and not connected to timer. Somethimes there is different diced display: only one, both or one from side. Also players dont see if roll doubles

doniliks (commented):
Timing is still a bit problematic in a sense that it takes a bit too long but we can adjust it later. If possible, we have to adjust current dice roll server part and adjust phase bot at the end of the dice roll, but upon calling resolve effects. I don’t merge it now since i want one other opinion as well

eles17 (commented):
For me looks fine after the danylos request is addressed

doniliks (approved):
looks nice

PR #361, opened 2026-06-20
ui refactoring

doniliks:
ui staff:
- smaller dices for left part
- different dice selection 

- added text info for non active players about active player actions: 
Screenshot 2026-06-21 at 01 11 01
Screenshot 2026-06-21 at 01 10 54

Screenshot 2026-06-21 at 01 11 18
Screenshot 2026-06-21 at 01 10 44

closes #347, #363

### DIFF
--- a/app/src/main/java/com/machikoro/client/ui/game/ui/Dice.kt
+++ b/app/src/main/java/com/machikoro/client/ui/game/ui/Dice.kt
@@ -22,13 +22,14 @@ import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.draw.alpha
 import androidx.compose.ui.graphics.Color
+import androidx.compose.ui.layout.ContentScale
 import androidx.compose.ui.res.painterResource
 import androidx.compose.ui.semantics.Role
 import androidx.compose.ui.semantics.contentDescription
-import androidx.compose.ui.semantics.role
 import androidx.compose.ui.semantics.semantics
 import androidx.compose.ui.text.font.FontWeight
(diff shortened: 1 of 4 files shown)


valentinaschiavon99 (dismissed):
One small fix. This PR is pure UI and introduces no protocol changes, so there are no mismatches with Server `main` (the ROLL_DICE/reroll/snapshot handling it relies on is unchanged here).

app/src/main/java/com/machikoro/client/ui/game/ui/Dice.kt
```diff
@@ -287,10 +369,11 @@ fun DiceSection(
                             onRollDice(chosen)
                         },
                         enabled = true,
-                        label = "Roll Dice",
+                        label = "Roll " + (if(chosen > 1) "Two " else "One ") + "Dice",
```
valentinaschiavon99: **"Roll One Dice" is grammatically off** — the singular of "dice" is "die", so the one-die case should read "Roll One Die". This fixes the original "Roll Dices" bug but introduces "One Dice". Suggested:

```kotlin
label = "Roll " + (if (chosen > 1) "Two Dice" else "One Die"),
```

Minor, but since we're already touching this label it's worth getting right.

valentinaschiavon99 (dismissed):
One wording fix on the dice label. The Column refactor on the result branch looks good — it groups the label and dice with proper spacing/centering.

valentinaschiavon99 (approved):
Wording fix applied ("Roll One Die" / "Roll Two Dice"). Column refactor is clean, builds locally, UI-only with no server/protocol changes. LGTM.

PR #364, opened 2026-06-22
330 add in game chat UI

SoupCube:
### Summary 
Added a UI to use the ingame chat during the game.

- A button at the bottom right of the game screen opens a semi transparent chat window
- tapping the X button in the top right of the chat or out side the chat window closes the chat
- user can enter messages in a input field and send them if the message isn't empty
- A non-persistent chat history is shown to the user
- the chat history automatically scrolls with new messages


valentinaschiavon99 (dismissed):
Thanks for this! The chat wiring matches the server contract (`/app/game.chat.send` + `{type:CHAT, sender, content, gameId}` on `/topic/game/{id}`). A few changes needed before merge — one is a guaranteed crash.

## Must fix — crashes the game screen on load

In `GameScreen.kt`:

```kotlin
currentPlayer = state.players.filter{it.isCurrentPlayer}[0].displayName,
```

This is evaluated on every recomposition (even with the chat closed), and `[0]` throws `IndexOutOfBoundsException` when:
- `state.players` is empty — the initial state during init, while `GameScreen` composes under `InitializationLoadingOverlay`, and
- no player has `isCurrentPlayer == true` yet — it's only set once `userId == localUserId` resolves.

Either case crashes the whole screen. Every other call site uses the safe idiom — please match it:

```kotlin
currentPlayer = state.players.firstOrNull { it.isCurrentPlayer }?.displayName ?: "",
```

## Should fix — message length mismatch with server

The server silently drops messages longer than 300 chars (`GameController.sendGameChat`, `maxLength = 300`). The input has no cap, so an over-length message just disappears with no feedback. Please add a client-side limit (e.g. cap `onValueChange` at 300 and/or show a counter) so the UX matches the server.

## Minor — please also clean up

1. Remove the 4 unused imports in `GameScreen.kt`: `Button`, `ButtonDefaults`, `DiceAnimationDisplay`, `PrimaryBeigeLight` (SonarCloud will flag these).
2. Unread badge counter: `LaunchedEffect(state.chatMessages.size) { unreadCount++ }` increments by 1 per change event regardless of how many messages arrived, and fires once on initial composition. Consider tracking the delta against the last-seen count instead.
3. Drop the redundant `"type":"CHAT"` field from the game-chat send body — `GameChatRequest` only has `gameId`/`message`.
4. `Divider` is deprecated in Material3 → use `HorizontalDivider`.
5. Add the missing trailing newline to `ChatInput.kt` and `ChatOverlay.kt`.
6. `messages.reversed()` allocates a new list every recomposition; since `reverseLayout = true` is already set, you can drop `.reversed()`.

The blocker is the crash. The rest is polish.

valentinaschiavon99 (commented):
## Code Review: 330 add in game chat UI

### Summary
Adds a floating in-game chat button with a slide-in overlay, message history, and a send input. The UI wiring is clean and the feature scope is well-contained.

---

### What is failing and why

**CI: `OkHttpWebSocketClientTest > sendChatMessageSendsStompFrame` — `org.json.JSONException` at line 3670**

The test fails because the PR removed `.put("type", "CHAT")` from the JSON body in `OkHttpWebSocketClient.kt`:

```diff
-    val body = JSONObject()
-        .put("type", "CHAT")
-        .put("message", message)
```

The existing unit test `sendChatMessageSendsStompFrame` asserts that the outgoing STOMP frame body contains `"type": "CHAT"`. Removing that field causes a `JSONException` when the test tries to read the missing key.

**Fix:** Either restore the `"type": "CHAT"` field in the JSON body, or update the test to match the new contract — but the server likely expects this field, so coordinate with the backend before dropping it. If the server no longer needs the type discriminator (because the endpoint is already typed), update the test to reflect the new expected payload.

**SonarCloud** is failing as a downstream consequence of the same broken build/test run.

---

### General review

**Correctness issues**

- **Missing `"type": "CHAT"` in the network payload** — as above. If the backend STOMP handler routes on this field, dropping it will silently fail at runtime, not just in tests.

- **Double-reversal in `ChatOverlay`** — `LazyColumn` is configured with `reverseLayout = true`, but then `messages.reversed()` is passed to `items(...)`. This reverses the list twice, so messages end up in the original (oldest-first) order but the list scrolls to the top instead of the bottom. Either use `reverseLayout = true` with `items(messages)` (no manual reverse), or use `reverseLayout = false` with `items(messages.reversed())`.

- **`_chatMessages` is a `mutableStateListOf` but exposed via `copy(chatMessages = _chatMessages)`** — `copy()` takes a snapshot of the list reference at that moment. Since `_chatMessages` is mutable and is passed directly, the state update is correct only by coincidence because Compose observes the `StateFlow`. Consider using `toList()` to be explicit: `.copy(chatMessages = _chatMessages.toList())`.

**Code quality**

- **Commented-out code left in** — `ChatInput.kt` has two commented-out lines (`//.padding(8.dp)` and `//Spacer(...)`). Remove before merging.

- **`modifier` parameter unused** — `ChatOverlay` accepts a `modifier: Modifier = Modifier` parameter but applies `Modifier.fillMaxSize()` hardcoded on the `AnimatedVisibility`, ignoring the parameter entirely. Either use it or remove it.

- **300-char message limit enforced only in the UI** — `ChatInput` blocks sending messages over 300 chars, but `GameScreenViewModel.sendChatMessage` does no length validation. If the UI check is bypassed (e.g. via direct ViewModel call or future refactor), oversized messages go straight to the server. Add the guard in the ViewModel too, or at minimum add a `maxLines`/`maxLength` constraint to the `OutlinedTextField`.

- **`currentPlayer` identification is fragile** — `state.players.firstOrNull { it.isCurrentPlayer }?.displayName` assumes `isCurrentPlayer` is set correctly on the player list; it will silently fall back to an empty string if not. Using `state.myUserId` to look up the display name would be more robust.

**Minor**

- Trailing newline missing on both new files (`ChatInput.kt`, `ChatOverlay.kt`) — most editors/linters will flag this.
- The `Badge { }` trailing space before `{` on the content lambda is a minor style inconsistency with the rest of the file.

valentinaschiavon99 (dismissed):
Thanks for this — the server contract is correct (send body now matches `GameChatRequest{gameId, message}`, receive/validation aligned with `origin/main`). A few changes before merge:

---

### 1. `parseInt(it.id)` can crash the game screen (please fix)

In `GameScreen.kt`:

```kotlin
currentPlayer = state.players.firstOrNull { parseInt(it.id) == state.myUserId }?.displayName ?: ""
```

`PlayerCoinState.id` is a `String`, and `java.lang.Integer.parseInt` throws `NumberFormatException` on any non-numeric/empty id. This runs on every recomposition, so a single bad id crashes the whole game screen. Use the null-safe Kotlin conversion and drop the `java.lang.Integer.parseInt` import:

```kotlin
currentPlayer = state.players.firstOrNull { it.id.toIntOrNull() == state.myUserId }?.displayName ?: ""
```

---

### 2. SnapshotStateList leaked into the StateFlow state (please fix)

In `GameScreenViewModel.kt`:

```kotlin
private val _chatMessages = mutableStateListOf<ChatMessageState>()
...
webSocketClient.chatMessages.collect { chat ->
    _chatMessages.add(chat)
    mutableState.update { it.copy(chatMessages = _chatMessages) }
}
```

`copy` receives the *same* list reference that was just mutated in place, so from the 2nd message on `newState.equals(oldState)` is `true` and `MutableStateFlow` dedupes — **it stops emitting**. The UI only keeps updating by accident because `state.chatMessages` is a Compose `SnapshotStateList` observed during composition. Any non-Compose collector won't see chat updates, and it breaks the immutability contract of `GameScreenState`.

Use a plain immutable list and emit a real copy:

```kotlin
// remove the _chatMessages field and the mutableStateListOf import
webSocketClient.chatMessages.collect { chat ->
    mutableState.update { it.copy(chatMessages = it.chatMessages + chat) }
}
```

`GameScreenState.chatMessages` stays `List<ChatMessageState>`, and `+ chat` produces a new list each time so the StateFlow emits normally.

---

### 3. Minor: self-message highlight couples displayName to username

`ChatOverlay` highlights `msg.sender == currentPlayer`, where `currentPlayer` is a `displayName` but `msg.sender` is the server's `principal.username`. They match today only because `displayName` is currently populated from `username`. Worth a short comment noting the assumption so it doesn't silently break if the two ever diverge.

valentinaschiavon99 (approved):
LGTM ✅ — all three review points are resolved:

1. **`parseInt` crash** — now `it.id.toIntOrNull() == state.myUserId`, and the `java.lang.Integer.parseInt` import is gone. No more `NumberFormatException` risk on recomposition.
2. **StateFlow dedupe / leaked `SnapshotStateList`** — replaced with `it.copy(chatMessages = it.chatMessages + chat)`, so a fresh list is emitted each message and `GameScreenState` stays immutable.
3. **displayName/username coupling** — documented with a comment on the self-message highlight.

Unused imports left over from the refactor were also cleaned up. Server contract matches `origin/main` (`GameChatRequest{gameId, message}` outbound, `CHAT`/`content`/`sender` inbound). Approving.

PR #367, opened 2026-06-23
refactor: dice images and padding updates

doniliks:
New dice images, added padding and correct text display
Screenshot 2026-06-23 at 14 11 13
Screenshot 2026-06-23 at 14 10 38


Ivona-Goranova (dismissed):
Small UI edge case: the fallback dice image should only be shown during `ROLL_DICE`.

Right now `DiceSection` is also used during `RESOLVE_EFFECTS`, so if `diceResult` is temporarily `null` there, the active player could see a standalone dice image even though they are not supposed to roll.

**Maybe we could change:**

else if (!isAnimating && state.diceResult == null) {

**to:**

else if (
    state.gamePhase == GamePhase.ROLL_DICE &&
    !isAnimating &&
    state.diceResult == null
) {

valentinaschiavon99 (dismissed):
Visuals look good and CI is green — this is a self-contained UI change with no server-protocol impact. A few cleanups before merge:

**1. Redundant `!isAnimating` checks**
- `Dice.kt:323` — `if(!state.isActivePlayer && !isAnimating)` sits inside the `state.diceResult != null` branch of the `when`, which is only reachable when `isAnimating` is already false (the `isAnimating ->` branch wins first). Drop the `&& !isAnimating`.
- `Dice.kt:361-364` — the new `else if` re-checks `!isAnimating`, but it's already inside the `if (state.isActivePlayer && … && !isAnimating)` block at line 341. Drop the redundant `!isAnimating` from the condition.

**2. `modifier` reuse on inner element (`Dice.kt:369`)**
The new static dice `Image` applies the `DiceSection`-level `modifier` param to an inner child. Use `Modifier.size(100.dp)` instead so the section's external modifier isn't silently duplicated onto a child.

**3. Magic offsets**
`offset(y = 5.dp)`, `offset(y = (-45).dp)`, `spacedBy((-20).dp)` are hand-tuned to the new art. Fine for now, but a short comment noting they're tuned to the dice assets would help the next person who touches them.

None of these are correctness bugs — happy to approve once the two redundant conditions are cleaned up.

valentinaschiavon99 (approved):
All requested changes addressed — redundant `!isAnimating` checks removed, inner Image now uses `Modifier`, and the tuned offsets are commented. CI is green and branch is up to date with main. LGTM 🎲

PR #370, opened 2026-06-27
369 display purchased card for all players

eles17:
## Summary

Displays successful purchase feedback to all players during the Buy or Build phase.

Previously, all clients received the purchase broadcast, but non-active players ignored it because they had no locally pending purchase. They continued seeing the waiting message instead of the purchased card.

## Changes

- Allow non-active players to apply purchase-success broadcasts.
- Preserve pending-purchase validation for the active player.
- Ensure only the active player ends the turn.
- Add regression coverage for non-active purchase feedback.

## Current flow:

1. Non-active player sees “Player X is deciding what card to buy.”
2. Active player selects and confirms a card.
3. Server broadcasts `PURCHASE_COMPLETED` to all players.
4. Both players display the purchased card for about 5 seconds.
5. Only the active player’s client calls `endTurn`.
6. The server advances the turn, and both screens continue to the next phase/player.


doniliks (approved):
looks fine, thanks

PR #371, opened 2026-06-27
ui: add coins animation

Ivona-Goranova:
## Summary

Closes #304  
Closes #305

This PR finalizes the resolving-effects coin animation flow and documents the server/client contract for resolving-effects data.

### Coin gain/loss animations

- Added coin transfer animation during `RESOLVE_EFFECTS`.
- Coins now visually move between affected players when one player pays another.
- Bank-to-player gains are animated for income effects.
- Player-to-bank losses are animated for pure deductions with no receiving player.
- Active transfer participants are highlighted:
  - green for coin gains
  - red for coin losses

### Server-authoritative final balances

- Coin animations are planned from the local resolving-effects preview.
- Visible coin balances remain bound to the authoritative server snapshot.
- The UI no longer treats predicted preview balances as final before the server confirms the resolved state.
- After animation completes, the client calls `resolveEffects`, and the final balances come from `payload.state.players[].coins`.

### Resolve-effects flow timing

- The resolving-effects phase no longer relies on a visible countdown for effect resolution.
- The active player’s client waits for coin animations to complete before sending `/app/game.resolveEffects`.
- A 20-second failsafe remains in the ViewModel in case animation completion does not fire.
- Existing Radio Tower reroll behavior remains respected: auto-resolution does not fire while a reroll decision is pending.

### Coin transfer planning

- Added a transfer planner that compares pre-resolution and previewed post-resolution balances.
- Handles:
  - one payer to multiple receivers
  - multiple payers to one receiver
  - bank-to-player gains
  - player-to-bank losses
  - mixed player transfers plus bank remainders
  - no-op balance changes

### Resolving-effects server contract documentation

- Added `docs/resolving-effects-server-contract.md`.
- Documented what the client currently receives from the server.
- Documented what the client currently derives locally.
- Confirmed server-side behavior:
  - `EFFECTS_RESOLVED.payload.coinDeltas` contains only non-zero net deltas.
  - `coinDeltas` keys are player DB IDs, not user IDs.
  - `payload.state.players[].coins` already contains final post-resolution balances.
  - no pre-resolution server event currently exposes triggered establishments, affected players, or transfer logs.
  - TV Station and Business Center target/swap details are not currently exposed.
  - red-card and Stadium payments are capped by available payer coins.
  - future transfer contracts should use explicit `"BANK"` instead of `null`.
- Updated `docs/websocket-protocol.md` to link the resolving-effects contract.

## Notes

The current implementation supports a best-effort client-side animation preview while keeping final balances server-authoritative. Fully authoritative per-card/per-transfer animations would require additional server data, documented in `docs/resolving-effects-server-contract.md`.

## Verification

- `./gradlew testDebugUnitTest --tests com.machikoro.client.ui.game.ui.coin_animation.CoinTransferPlannerTest`
- `./gradlew testDebugUnitTest`
- `./gradlew lint`


valentinaschiavon99 (dismissed):
Nice work — the planner is clean and well-tested, balances stay server-authoritative, and the ID keying is consistent throughout. A couple of things to fix before merge.

## 1. (blocking) 20-second stall when a card triggers but no coins move

The active player's `resolveEffects` is now driven by the coin overlay's `onFinished`. But the overlay only renders when `coinTransfers` is non-empty — `coinAnimationFinished` is initialized to `coinTransfers.isEmpty()`, and the overlay is gated on `!coinAnimationFinished`. Meanwhile `resolveEffectsDwellMillis()` only returns `0L` when `triggeredCount == 0`; for any `triggeredCount > 0` it returns the 20s failsafe.

So when a card triggers but the preview balances don't change, nothing ever calls `finishResolveEffectsAnimation()`, and the turn hangs on `RESOLVE_EFFECTS` for the **full 20 seconds** before the failsafe fires (previously ~2–4s).

Concrete triggers:
- Active player has **0 coins** and rolls into opponents' red cards (Cafe/Restaurant) — the documented red-card cap makes the payment `0`, so `coinTransfers` is empty.
- Any zero-net-delta trigger, e.g. Business Center (income `0`).

**Fix** — when there's nothing to animate, signal completion immediately so the active player advances promptly. In `GameScreen.kt`, in the existing reset effect:

```kotlin
LaunchedEffect(animationKey) {
    coinAnimationFinished = coinTransfers.isEmpty()
    activeCoinTransfer = null
    if (state.gamePhase == GamePhase.RESOLVE_EFFECTS && coinTransfers.isEmpty()) {
        onResolveEffectsAnimationFinished()
    }
}
```

This is safe and idempotent: `finishResolveEffectsAnimation()` already no-ops for non-active players (`shouldAutoResolveEffects`) and when the job is already inactive (`!pendingJob.isActive`). It restores the fast advance for the no-coin-movement case while keeping the 20s value as a true failsafe.

(Alternative: make `resolveEffectsDwellMillis()` return `0L` when the planned transfer set is empty — but the VM doesn't currently know about `coinTransfers`, so the UI-side fix above is the smaller change.)

## 2. Stale comment referencing a deleted effect

`Dice.kt` removes the `LaunchedEffect(state.isRolling)` block, but the surviving comment above the constants still says:

> The active player's spin also ends the instant the server confirms the result (see the isRolling effect below) ...

That effect no longer exists — the spin now always runs the fixed `DICE_ANIMATION_DURATION_MS` (1500ms).

**Fix** — drop the now-false clause so the comment matches behavior:

```kotlin
// One brief roll animation for everyone, so the reveal isn't held behind a long fixed
// delay and active/non-active players stay in lockstep (matters during a Radio Tower
// reroll where both spin off the same roll tick).
private const val DICE_ANIMATION_DURATION_MS = 1500L
```

## 3. (nit) Missing trailing newline

`CoinTransferUi.kt` ends without a trailing newline (`\ No newline at end of file`). Add one.

valentinaschiavon99 (dismissed):
Nice work — the core mechanism is solid: the `finishResolveEffectsAnimation` guard (`isActive` + `shouldAutoResolveEffects` + nulling the job) correctly prevents a double `resolveEffects` against the dwell-0 / failsafe paths, the planner conserves coins and its tests pass, and snapshot ids line up across `PlayerCoinField` and `PlayersTopBar`. No correctness bugs found. A few cleanup items before merge:

### 1. Orphaned comment describing deleted behavior — `Dice.kt`

The PR removes the `LaunchedEffect(state.isRolling)` that stopped the active player's dice spin early on server confirm, but leaves its two-line comment (`// Stop animation for active player as soon as server confirms the result…`) now sitting directly above `Column(`. It documents code that no longer exists and will mislead the next reader.

This is also a real behavior change worth a line in the PR description: the active player's dice now always spins the full `DICE_ANIMATION_DURATION_MS` rather than ending the instant the result lands.

### 2. The two renamed ViewModel tests are now functionally identical

`resolveEffectsSendsWhenSingleCardAnimationFinishes` and `…MultipleCardAnimationsFinish` differ only in card quantity (1 vs 2) and `advanceTimeBy(1_999L)` vs `advanceTimeBy(3_999L)`. Since the dwell is now a flat 20s failsafe and resolution is driven by `finishResolveEffectsAnimation()`, neither the quantity nor those time advances exercise distinct behavior anymore — both assert "doesn't fire on its own, fires on finish." The `1_999`/`3_999` values are leftovers from the old graduated-dwell logic. One test with a clearer name would cover it.

### 3. Failsafe dwell raised 6s → 20s — confirm intent

Previously the active player auto-advanced within at most 6s. Now, if the overlay's `onFinished` never fires (e.g. `positionsReady` in `CoinTransferOverlay` never becomes true because a transfer participant's badge isn't laid out / doesn't report `onGloballyPositioned`), the player stares at `RESOLVE_EFFECTS` for a full 20s before the failsafe kicks in. Layout normally reports positions so this is unlikely to trigger, but the worst-case stall tripled — consider a shorter failsafe or logging if it expires.

(The preview-vs-authoritative mismatch is real but explicitly documented as a known best-effort limitation in `docs/resolving-effects-server-contract.md`, so not flagging it.)

valentinaschiavon99 (approved):
Re-reviewed at 4ba6a6c5 — all three points from my earlier review are addressed cleanly:

- Stale `Dice.kt` early-stop comment removed.
- Duplicate ViewModel test dropped; survivor renamed and the magic time advances replaced with `dwell - 1` so they track the constant.
- Failsafe dwell reduced 20s → 6s with a clarifying comment.

No new issues — the failsafe change is a single constant and the test logic is preserved (no auto-send before the dwell, sends on `finishResolveEffectsAnimation`). LGTM.

PR #372, opened 2026-06-27
refactor: Refactored the navigation to the global leaderboard and back

Moritz2k1:
Fulfilled the things that were wanted in the issue


valentinaschiavon99 (approved):
Reviewed the navigation refactor. Clean change: routes the two leaderboard navigations (Home → Ranking, Winner → View Leaderboard) through NavigationViewModel.navigateTo(...) and adds a popUpTo exception so the leaderboard behaves as an overlay, preserving the back stack so the back button returns to Home/Winner. Functionally equivalent to the previous direct navController.navigate calls, plus the dedup guard. No server contract impact — the network/leaderboard layer is untouched. Test matches the existing pattern. LGTM.

Note: filed a follow-up issue for a pre-existing latent race (state-driven NavigateTo(Home) can bounce the user off the leaderboard) — not introduced by this PR.

PR #374, opened 2026-06-27
fix: prevent leaderboard navigation from being bounced back to Home (#373)

valentinaschiavon99:
## Summary

Fixes the latent navigation race described in #373: the leaderboard could be unexpectedly bounced back to **Home** by a competing state-driven navigation event.

The leaderboard is opened by an explicit user action (tapping **View Leaderboard** on the Winner screen, or **Ranking** on Home), but it layers on top of the state-driven flow rather than being derived from app state. Previously a competing `updateNavigationBasedOnState` call could compute `Home` and navigate away from it:

- On the Winner screen, `onClearGameState()` runs before navigating to the leaderboard. Clearing finished-game state triggers `updateNavigationBasedOnState`, which (logged in, no active/finished game) resolves to `Home`.
- `clearLastNavigation()` fires on every `OnDestinationChangedListener` callback, so the `lastNavigation` dedup guard reopens right after the leaderboard navigation lands.
- A stale `NavigateTo(Home)` can therefore land **after** the leaderboard navigation and bounce the user back.
- The same mechanism let a **server snapshot arriving while the leaderboard is open** pull the user back to Home.

## Approach

This implements the recommended "overlay route" direction from the issue.

`NavigationViewModel` now tracks the current destination and treats `Leaderboard` as an **overlay route**. While an overlay route is the current destination, `updateNavigationBasedOnState` does not navigate away from it, so background/state-driven `NavigateTo` events can no longer override an explicit user navigation. A forced logout remains an exception, since unauthenticated users must always return to `Main`.

## Changes

- **`NavigationViewModel.kt`**
  - Tracks `currentRoute`, updated via the new `onDestinationChanged(route)` (which also clears `lastNavigation`, preserving prior behavior).
  - Added an `overlayRoutes` set (currently `Leaderboard`).
  - `updateNavigationBasedOnState` skips navigation when an overlay route is in front, unless the target is `Main` (logout).
- **`AppRoot.kt`** — the `OnDestinationChangedListener` now calls `onDestinationChanged(destination.route)` instead of `clearLastNavigation()`, so the ViewModel knows which route is in front.
- **`docs/navigation.md`** — documents the `Leaderboard` route and the overlay-route behavior.

## Tests

Added unit tests in `NavigationViewModelTest`:
- `stateUpdateWhileOnLeaderboardDoesNotNavigateAway` — a state update that would route to Home emits nothing while on the leaderboard.
- `logoutWhileOnLeaderboardStillNavigatesToMain` — a forced logout still wins over the overlay.
- `stateUpdateNavigatesNormallyWhenNotOnOverlay` — normal state-driven routing is unaffected once off the overlay.
- `onDestinationChangedTracksRouteAndClearsLastNavigation` — verifies route tracking and dedup-cache clearing.

### DIFF
--- a/app/src/main/java/com/machikoro/client/ui/AppRoot.kt
+++ b/app/src/main/java/com/machikoro/client/ui/AppRoot.kt
@@ -106,8 +106,8 @@ fun AppRoot(
     val showConnectionBanner = currentRoute != null && currentRoute != AppRoute.Main.route
 
     DisposableEffect(navController) {
-        val listener = NavController.OnDestinationChangedListener { _, _, _ ->
-            navigationViewModel.clearLastNavigation()
+        val listener = NavController.OnDestinationChangedListener { _, destination, _ ->
+            navigationViewModel.onDestinationChanged(destination.route)
         }
         navController.addOnDestinationChangedListener(listener)
         onDispose {
--- a/app/src/main/java/com/machikoro/client/ui/AppRoot.kt
+++ b/app/src/main/java/com/machikoro/client/ui/AppRoot.kt
@@ -128,7 +128,11 @@ fun AppRoot(
         gameScreenState,
         startScreenState,
         lobbyCode,
-        navigationUiState.showLobbyScreen
+        navigationUiState.showLobbyScreen,
+        // Re-evaluate when the destination changes so that leaving an overlay
+        // route (e.g. Back from Leaderboard) recalculates the state-driven
+        // target and doesn't strand the user on a screen with stale data (#373).
+        currentRoute,
     ) {
         navigationViewModel.updateNavigationBasedOnState(
             gameScreenState = gameScreenState,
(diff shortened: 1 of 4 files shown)


eles17 (dismissed):
Thanks for the fix. Please address this before merging:

When Winner → Leaderboard clears the game state, the resulting navigation to Home is suppressed while Leaderboard is active. After pressing Back, the app returns to Winner, but navigation is not recalculated because the current route is not a `LaunchedEffect` dependency. This can leave users on a Winner screen with cleared/empty game data.

Please either:
- re-evaluate state-driven navigation when leaving an overlay, or
- avoid clearing the finished-game state when opening Leaderboard.

Please also add a test covering: **Winner → Leaderboard → Back**.

eles17 (approved):
Approved 👍 
The Winner → Leaderboard → Back issue is now correctly handled, with regression coverage included.

PR #376, opened 2026-06-28
366 refactor other players card view

doniliks:
Changes UI - now players can click on opponets name and see their cards, if more that one oppenent, can be switched between them

Screenshot 2026-06-28 at 02 31 49
Screenshot 2026-06-28 at 02 31 26


eles17 (changes requested):
Could you filter and sort the cards before displaying them?

The dialog currently renders the raw `cards` list, so cards with `quantity = 0` may appear even though the player does not own them. Please use:

```kotlin
cards.visibleInDisplayOrder()
```

before `chunked(4)`. This also preserves the intended card ordering.

eles17 (approved):
Thanks for correcting sorting. Apporved.

PR #377, opened 2026-06-28
Delete chatMessages when Game ID changes

SoupCube:
### Summary
Added a line to `activeGameId.collect` to now also empty the message list whenever the gameID changes.
Added a test to assert this behavior works as intended.


valentinaschiavon99 (approved):
## Approved ✅

Clean, well-scoped fix for #368. Clearing `chatMessages` inside the `activeGameId` collector is the right hook: it resets the chat both when a new game starts and when dropping back to the lobby, which matches the issue's intent. Verified that `activeGameId` only goes `null` on an intentional teardown (`clearGameState`/`resetLobbyState`), not on a transient reconnect (the id is retained and reused to resubscribe), so chat is not wiped when the app just backgrounds mid-game.

### Verified locally
- `:app:compileDebugUnitTestKotlin` → BUILD SUCCESSFUL
- `chatMessagesAreClearedInAFreshGame` → passes
- Purely client-side state change — no protocol/network surface touched, so no mismatch with the Server.

### Note
Pushed a tiny follow-up commit removing three unused imports in the test (`launchIn`, `onEach`, and the `private` `FakeWebSocketFactory`) that SonarCloud would have flagged as code smells.

LGTM 🚀

PR #379, opened 2026-06-28
378 dice cards purchased card

doniliks:
Dice: added dice image if only one dice
GamePhaseBanner: changed opacity like on sketches
Marketplace: added title on form of phase field
ResolveEffects: changed coin to -/+ inside coin and added scrollable column, adjusted placing, 
and general placing changes

### DIFF
--- a/app/src/main/java/com/machikoro/client/ui/game/GameScreen.kt
+++ b/app/src/main/java/com/machikoro/client/ui/game/GameScreen.kt
@@ -472,7 +472,8 @@ fun GameScreen(
                                 horizontalAlignment = Alignment.CenterHorizontally,
                                 verticalArrangement = Arrangement.spacedBy(12.dp)
                             ) {
-                                ResolvingEffectsView(state = state)
+                                ResolvingEffectsView(state = state,
+                                    modifier = Modifier.offset(y = -50.dp))
 
                                 if (
                                     showRadioTowerReroll &&
@@ -540,7 +541,8 @@ fun GameScreen(
                         val turnFlowLabel = state.turnFlowActionLabel()
                         if (
                             state.isActivePlayer &&
-                            state.gameStatus == GameStatus.IN_PROGRESS
+                            state.gameStatus == GameStatus.IN_PROGRESS &&
+                            state.purchaseState != PurchaseState.SUCCESS
                         ) {
                             turnFlowLabel?.let {
                                 ActionButton(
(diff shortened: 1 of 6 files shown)


valentinaschiavon99 (dismissed):
The button-hiding change on `purchaseState == SUCCESS` is safe: that button was already disabled on SUCCESS and turn advancement is server-driven.

A few non-blocking cleanups before merge:

**1. Remove unused imports — `Marketplace.kt`**
`PlayerCoinState` and `PrimaryOrange` are imported but never used. Delete:
```kotlin
import com.machikoro.client.domain.model.state.PlayerCoinState
import com.machikoro.client.ui.theme.PrimaryOrange
```

**2. Fix misleading indentation — `ResolvingEffectsView.kt`**
`OutcomeStack` is indented 4 spaces as if nested inside `PlayerOutcomeColumn`. It's actually a top-level function (compiles fine) — please de-indent it to column 0 to line up with the other top-level `@Composable` functions.

**3. Honor the passed `modifier` in the empty state — `ResolvingEffectsView.kt`**
`GameScreen` passes `Modifier.offset(y = -50.dp)`, but the "No establishments triggered" branch ignores it and hard-codes `offset(y = -80.dp)`, so placement differs depending on whether effects triggered. Either apply the incoming `modifier` to the `BasicText`, or, if `-80.dp` is intentional for this state, add a comment noting it deliberately overrides the caller's offset.

valentinaschiavon99 (approved):
Looks good overall — nice UI polish. A few things worth a follow-up:

- **Double offset on empty effects state**: `ResolvingEffectsView` now receives `Modifier.offset(y = -50.dp)` from the call site and then chains another `.offset(y = -35.dp)` internally for the "No establishments triggered" text. The two compose to -85 dp, which likely pushes the label off-screen. One of them should be removed.
- **Horizontal scroll removed from `TriggeredEffectsBoard`**: the outer `Row` lost `horizontalScroll` but each `PlayerOutcomeColumn` is 155 dp wide — 4 players = 620 dp, wider than any phone. The per-column vertical scroll doesn't replace row-level horizontal overflow. Worth checking on a 4-player game.
- **`CoinForEffects` typography mismatch**: `fontSize = 20.sp` is set on a `labelSmall` style whose line-height is 16 sp — the glyph is taller than its container and may clip on real devices.
- **Two unused imports in `Marketplace.kt`**: `PlayerCoinState` and `PrimaryOrange` were added but never used.

PR #382, opened 2026-06-28
fix(#380): negotiate and emit STOMP heartbeats

valentinaschiavon99:
## What & why

The client advertised `heart-beat: 0,0` in the STOMP CONNECT frame, which **disables** heartbeats. Server PR [#426](https://github.com/SE2-Machi-Koro/Server/pull/426) enables broker heartbeats, but heartbeats are negotiated — with the client opting out, negotiation still yielded none, so idle WebSocket connections kept being reaped by the platform proxy (Railway's edge) during a player's think-time. That was the mid-game lag / "connection timeout" players hit.

Simply flipping the header isn't enough: once the client advertises a non-zero send interval, the server **expects** client heartbeats and will drop the client if it doesn't actually send them. So this also implements real heartbeat emission.

Closes #380. Pairs with server #426.

## Changes

`OkHttpWebSocketClient`:
- CONNECT now advertises `heart-beat: 10000,10000` (matches the server's `HEARTBEAT_INTERVAL_MS`).
- On `CONNECTED`, read the negotiated `heart-beat` header and start a periodic emitter that sends a lone LF (`\n`, the STOMP heartbeat) every negotiated interval. Interval is computed per STOMP 1.2: `max(ourSend, serverRequestedReceive)`; emission is skipped when either side opts out.
- The emitter is cancelled when the socket goes away (`clearSocket` covers failure/close; `disconnect` covers the client-initiated path), so reconnects don't stack jobs.
- Inbound heartbeats need no new handling: `StompFrame.parseFrames` already trims lone newlines.

## Testing

New unit tests in `OkHttpWebSocketClientTest` (driven on the test scheduler via `backgroundScope` + `advanceTimeBy`):
- `connectFrameAdvertisesHeartbeat` — CONNECT contains `heart-beat:10000,10000`.
- `connectedNegotiatesAndEmitsHeartbeats` — a lone-LF heartbeat is emitted one interval after CONNECTED.
- `heartbeatsStopAfterDisconnect` — no heartbeats after `disconnect()`.
- `serverOptingOutDisablesHeartbeatEmission` — server `heart-beat: 0,0` ⇒ no emission.

`./gradlew :app:testDebugUnitTest` — **BUILD SUCCESSFUL** (185 tests, 0 failures).

## Notes

- Keep `CLIENT_HEARTBEAT_INTERVAL_MS` (10s) in sync with the server constant.
- Receive-side liveness detection (reconnect if the server goes silent) is left to the existing auto-reconnect on socket close/failure; can be added later if needed.

### DIFF
--- a/app/src/main/java/com/machikoro/client/network/websocket/OkHttpWebSocketClient.kt
+++ b/app/src/main/java/com/machikoro/client/network/websocket/OkHttpWebSocketClient.kt
@@ -233,8 +233,9 @@ class OkHttpWebSocketClient(
     override fun disconnect() {
         intentionalDisconnect = true
         cancelReconnect()
-        stopHeartbeat()
         val currentSocket = synchronized(this) {
+            heartbeatJob?.cancel()
+            heartbeatJob = null
             val socket = webSocket
             webSocket = null
             socket
@@ -1406,8 +1407,9 @@ class OkHttpWebSocketClient(
     }
 
     private fun clearSocket() {
-        stopHeartbeat()
         synchronized(this) {
+            heartbeatJob?.cancel()
+            heartbeatJob = null
             webSocket = null
             subscribedGameId = null
             stompSessionId = null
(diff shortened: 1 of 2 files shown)


eles17 (approved):
Thanks. Approved.

PR #384, opened 2026-06-28
ci: add GitHub Actions workflow for itch.io APK deployment

valentinaschiavon99:
## Summary

This PR implements a Continuous Deployment pipeline that automatically builds a signed Android APK and publishes it to the project's itch.io page whenever a commit is pushed to \`main\` or a new GitHub Release is published.

### New: \`.github/workflows/deploy-to-itch.yml\`

A dedicated CD workflow with the following stages:

1. **Checkout & JDK 21 setup** — mirrors the existing CI configuration.
2. **Decode Keystore** — base64-decodes \`KEYSTORE_FILE\` via an \`env:\` variable (not inline secret expansion) to avoid secret leakage in shell logs.
3. **Build signed Release APK** — runs \`./gradlew assembleRelease\` with signing credentials injected through environment variables.
4. **Install Butler (pinned)** — downloads butler \`v15.27.0\` from the official GitHub release URL (not the mutable \`LATEST\` endpoint), enforces HTTPS with \`--proto '=https' --tlsv1.2\`, and verifies the SHA256 checksum before execution.
5. **Deploy to itch.io** — pushes the APK to itch.io on the \`android\` channel using \`BUTLER_API_KEY\`. \`ITCH_USER\` and \`ITCH_GAME\` are passed via \`env:\` to avoid direct secret expansion in the \`run\` command.

### Updated: \`app/build.gradle.kts\`

Added a \`signingConfigs.release\` block that reads \`KEYSTORE_PATH\`, \`KEYSTORE_PASSWORD\`, \`KEY_ALIAS\`, and \`KEY_PASSWORD\` from environment variables. Previously the signing secrets were passed to Gradle but never consumed, resulting in an unsigned release APK.

### Updated: \`.github/workflows/ci.yml\`

Applied the same signing and keystore handling improvements to the existing release artifact upload job.

## Required repository secrets

| Secret | Purpose |
|---|---|
| \`BUTLER_CREDENTIALS\` | itch.io API key (already configured) |
| \`ITCH_USER\` | itch.io username |
| \`ITCH_GAME\` | itch.io game slug |
| \`KEYSTORE_FILE\` | Base64-encoded Android keystore |
| \`KEYSTORE_PASSWORD\` | Keystore password |
| \`KEY_ALIAS\` | Key alias |
| \`KEY_PASSWORD\` | Key password |

## Trigger conditions

| Event | Triggers deploy? |
|---|---|
| Push to \`main\` | Yes |
| GitHub Release published | Yes |
| Push to other branches | No |
| Pull requests | No |

## Security notes

- No secrets are expanded inline in \`run:\` blocks — all are passed through \`env:\` and referenced as shell variables.
- Butler binary is pinned to \`v15.27.0\` with SHA256 verification; the workflow fails before execution if the checksum doesn't match.
- HTTPS is strictly enforced on the butler download (\`--proto '=https' --tlsv1.2\`).
- No \`.aab\` (Android App Bundle) is produced or uploaded — itch.io requires a standard \`.apk\`.

Closes #383

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -66,7 +66,9 @@ jobs:
         run: chmod +x gradlew
 
       - name: Decode Keystore
-        run: echo "${{ secrets.KEYSTORE_FILE }}" | base64 -d > keystore.jks
+        env:
+          KEYSTORE_FILE: ${{ secrets.KEYSTORE_FILE }}
+        run: echo "$KEYSTORE_FILE" | base64 -d > keystore.jks
 
       - name: Build Release APK
         run: ./gradlew assembleRelease
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -65,10 +65,13 @@ jobs:
       - name: Grant execute permission for gradlew
         run: chmod +x gradlew
 
+      - name: Decode Keystore
+        run: echo "${{ secrets.KEYSTORE_FILE }}" | base64 -d > keystore.jks
+
       - name: Build Release APK
         run: ./gradlew assembleRelease
         env:
-          KEYSTORE_FILE: ${{ secrets.KEYSTORE_FILE }}
+          KEYSTORE_PATH: ${{ github.workspace }}/keystore.jks
           KEYSTORE_PASSWORD: ${{ secrets.KEYSTORE_PASSWORD }}
           KEY_ALIAS: ${{ secrets.KEY_ALIAS }}
           KEY_PASSWORD: ${{ secrets.KEY_PASSWORD }}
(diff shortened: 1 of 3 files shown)


eles17 (dismissed):
The workflow is not ready to merge yet:

1. `assembleRelease` currently produces an unsigned APK. The signing secrets are passed as environment variables, but `app/build.gradle.kts` has no release signing configuration, so they are unused.
2. Pin `manleydev/butler-publish-itchio-action` to a full commit SHA instead of mutable `@master`, especially because it receives deployment credentials.
3. Please verify a successful build and test deployment before merging. SonarCloud is currently failing its security quality gate.

eles17 (dismissed):
Thanks for addressing the previous feedback. The signing configuration now looks correct, but one security concern remains:

Butler is downloaded from the mutable `LATEST` URL without checksum verification and then executed with `BUTLER_API_KEY`. Please pin a specific Butler version and verify its checksum, or use a setup action pinned to a full commit SHA.

Please also verify a successful signed APK build and test deployment before merging.

eles17 (approved):
Thanks for addressing requested changes. Approved.

PR #389, opened 2026-06-28
fix: surface private domain errors from /user/queue/errors (#381)

valentinaschiavon99:
Closes #381

## Context

Server PR [#428](https://github.com/SE2-Machi-Koro/Server/pull/428) fixed the global WebSocket exception handler so domain rejections (e.g. `NOT_YOUR_TURN`) are actually delivered to the caller on `/user/queue/errors` instead of being logged server-side and silently dropped.

The client already subscribes to `/user/queue/errors` (subscription id `user-errors`), but it did not parse or surface what the server now sends, so generic domain errors were silently ignored on the client even after the server fix.

## Problem — payload-shape mismatch

The server sends a bare `WebSocketErrorDto` with **top-level** fields and **no `type`**:

```json
{ "code": "NOT_YOUR_TURN", "message": "It is not your turn", "timestamp": 1714000000000, "context": {} }
```

But the MESSAGE dispatch mostly assumed the broadcast envelope shape (`type == "ERROR"`, nested `payload.code` / `payload.message`):

- `WsErrorParser.parse` read `payload.errorCode` / `payload.code` and `content` / `payload.message` — not the top-level fields, so a `NOT_YOUR_TURN` frame parsed as `UNKNOWN` with a generic message.
- `handleLobbyError` / `parsePurchaseFailure` require `type == "ERROR"`, so they skipped the raw DTO entirely.
- Only `handleAccusationError` read top-level `code`, and only for `INVALID_ACCUSATION`.

## Changes

1. **`WsErrorParser.parse`** now also reads **top-level** `code` and `message` (the `WebSocketErrorDto` contract), while staying backward-compatible with the nested-`payload` broadcast shape.
2. **New sink `domainErrors: SharedFlow<ClientError.WebSocket>`** on `WebSocketClient` / `OkHttpWebSocketClient`, fed by a new `handlePrivateError` that matches frames with **no `type`** and a top-level `code`. `INVALID_ACCUSATION` is excluded because it already has a dedicated `accusationErrors` flow — this avoids double-surfacing.
3. **Routing precedence**: `handlePrivateError` only fires for envelope-less frames (`type` blank), so private error frames are never mis-handled by the topic/broadcast handlers, and broadcast envelopes never leak into `domainErrors`.
4. **`GameScreenViewModel`** exposes `domainErrors` as a pass-through; **`MainActivity`** collects it and shows the server's message as a toast.

## Tests

- `WsErrorParserTest`: parsing a bare `WebSocketErrorDto` yields `serverCode = NOT_YOUR_TURN`, `message = "It is not your turn"`.
- `OkHttpWebSocketClientTest`:
  - `privateDomainErrorFrameEmitsDomainError` — a `NOT_YOUR_TURN` frame is surfaced on `domainErrors`.
  - `invalidAccusationDoesNotEmitDomainError` — `INVALID_ACCUSATION` is **not** surfaced on `domainErrors` (still handled by `accusationErrors`).
- Updated the other `WebSocketClient` test doubles (`WebSocketClientTest`, `HomeScreenViewModelTest`) to implement the new member.

All websocket and home-VM unit tests pass locally (`./gradlew testDebugUnitTest`).

## Acceptance criteria

- [x] Ending a turn out of order yields a parsed error with `serverCode = NOT_YOUR_TURN`, `message = "It is not your turn"`, surfaced to the user.
- [x] Existing `INVALID_ACCUSATION` and lobby-join error handling unchanged.

## Design note

The issue flagged an alternative: have the **server** wrap the error in a `WebSocketMessage` envelope (`type=ERROR`, `payload={code,message,context}`) to match the broadcast shape, avoiding a client parser change. The docs repo treats `/user/queue/errors` as a `WebSocketErrorDto`, so this PR adapts the client. If the team prefers a uniform envelope, that can be done server-side instead.

### DIFF
--- a/app/src/main/java/com/machikoro/client/MainActivity.kt
+++ b/app/src/main/java/com/machikoro/client/MainActivity.kt
@@ -227,6 +227,14 @@ class MainActivity : ComponentActivity() {
                 }
             }
 
+            // Generic domain rejections on /user/queue/errors (e.g. NOT_YOUR_TURN,
+            // server #428): surface the server's message as a toast.
+            LaunchedEffect(Unit) {
+                gameScreenViewModel.domainErrors.collect { error ->
+                    Toast.makeText(context, error.userMessage, Toast.LENGTH_LONG).show()
+                }
+            }
+
             // Debug End-game (#191): surface End-game button failures as a snackbar.
             LaunchedEffect(Unit) {
                 gameScreenViewModel.debugEndGameErrors.collect { message ->
--- a/app/src/main/java/com/machikoro/client/network/websocket/OkHttpWebSocketClient.kt
+++ b/app/src/main/java/com/machikoro/client/network/websocket/OkHttpWebSocketClient.kt
@@ -121,6 +121,9 @@ class OkHttpWebSocketClient(
     override val accusationErrors: SharedFlow<String>
         get() = mutableAccusationErrors.asSharedFlow()
 
+    override val domainErrors: SharedFlow<ClientError.WebSocket>
+        get() = mutableDomainErrors.asSharedFlow()
+
     override val chatMessages: SharedFlow<ChatMessageState>
         get() = mutableChatMessages.asSharedFlow()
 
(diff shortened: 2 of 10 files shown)


eles17 (approved):
LGTM. Approved

PR #392, opened 2026-06-28
ci: fix butler binary path in itch.io deploy

valentinaschiavon99:
## Problem

The `Deploy to itch.io` workflow fails at the **Install Butler** step:

```
chmod: cannot access 'butler': No such file or directory
Error: Process completed with exit code 1.
```

`butler-linux-amd64.zip` (v15.27.0) extracts the binary into a `linux-amd64/` subdirectory (`inflating: linux-amd64/butler`), but the step ran `chmod +x butler` against the working directory, where no `butler` exists. The later `./butler push` would have failed the same way.

## Fix

- `chmod +x ./linux-amd64/butler` (the real path)
- add `linux-amd64/` to `$GITHUB_PATH` so the deploy step finds `butler` on PATH
- call `butler push` instead of `./butler push`

## Note

This workflow also builds the signed release APK and already passes `KEYSTORE_FILE` correctly, so it benefits from the signing secrets fixed alongside #390.

Closes #393


eles17 (approved):
LGTM

PR #394, opened 2026-06-28
fix: Fixed game continuation

Moritz2k1:
Hope this works and fixes the issue. All tests pass.

### DIFF
--- a/app/src/main/java/com/machikoro/client/network/websocket/OkHttpWebSocketClient.kt
+++ b/app/src/main/java/com/machikoro/client/network/websocket/OkHttpWebSocketClient.kt
@@ -1070,18 +1070,23 @@ class OkHttpWebSocketClient(
         updateLobbyHostOwnership(resolveHostUserId(state, game))
 
         parseGameStatus(game.optString("status"))?.let { mutableGameStatus.value = it }
-        parseTurnPhase(game.optString("turnPhase"))?.let { mutableGamePhase.value = it }
+        val snapshotPhase = parseTurnPhase(game.optString("turnPhase"))
+        snapshotPhase?.let { mutableGamePhase.value = it }
         game.optIntOrNull("roundNumber")?.let { mutableRoundNumber.value = it }
-        // The snapshot persists only the dice total (lastDiceRoll), not the
-        // individual dice. Surface it as a single-element list so the last roll
-        // still shows on reconnect — but do NOT clobber a richer per-die result we
-        // already hold for the same roll (it sums to this total). Otherwise a
-        // routine snapshot collapses [x, y] into a single generic die mid-turn,
-        // which is the inconsistent "one vs two dice" display and also hides doubles.
-        game.optIntOrNull("lastDiceRoll")?.let { total ->
-            val current = mutableDiceResult.value
-            if (current == null || current.sum() != total) {
-                mutableDiceResult.value = listOf(total)
+        // In ROLL_DICE phase no dice have been rolled yet for this turn — lastDiceRoll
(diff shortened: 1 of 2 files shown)


valentinaschiavon99 (changes requested):
### Issues Requiring Changes

**1. Insufficient Test Coverage**

The added test only validates that ROLL_DICE snapshots clear diceResult in isolation. This does not confirm:
- Whether the Roll button actually becomes visible after reconnect (the observable bug)
- Whether other phase transitions work correctly
- Whether the fix addresses the root cause of issue #375

Request: Add tests for these scenarios:

a) Reconnect during ROLL_DICE with stale dice - verify Roll button is now clickable
b) RESOLVE_EFFECTS to BUY_OR_BUILD transition - confirm diceResult persists correctly
c) Non-ROLL_DICE snapshots - verify the else branch still restores lastDiceRoll

**2. Issue #375 Verification Missing**

PR description states "Hope this works and fixes the issue" but provides no evidence. Issue #375 specifically reports: "game does not continue with buying phase" after RESOLVE_EFFECTS phase.

Request: Clarify whether this PR was tested against issue #375's reproduction steps. If the root cause is different, the issue may remain unresolved despite this fix.

### Why This Matters

Issue #375 describes a game stuck in BUY_OR_BUILD phase. This PR only handles ROLL_DICE phase. Without comprehensive test coverage and validation, we cannot confirm whether:
- The stale dice is actually the root cause of #375
- Other state fields have similar persistence problems  
- The game can successfully transition through all phases after reconnect

### Recommendation

Approve the code changes, but add test coverage demonstrating this fix resolves the actual reported issue.

valentinaschiavon99 (approved):
## Approved — verified against server `main`

I re-reviewed this against the current server `main` branch and found **no bugs, errors, or mismatches**. The fix is correct and consistent with server semantics.

### Server-side verification
The fix rests on the assumption that a ROLL_DICE snapshot must not carry a usable dice result. The server **guarantees** this invariant — `turnPhase == ROLL_DICE` always implies `lastDiceRoll == null`:

- `GameDao.create` and `GameDao.advanceTurn` both set `turnPhase = ROLL_DICE` together with `lastDiceRoll = null` (and `lastDiceCount = null`).
- `tryRecordDiceRoll` is the **only** transition out of ROLL_DICE, and it sets `lastDiceRoll` while moving to `RESOLVE_EFFECTS` atomically.
- `DiceService.rollDice` additionally guards on `phase == ROLL_DICE && lastDiceRoll == null`.

So clearing `diceResult` on a ROLL_DICE snapshot is exactly correct, and the `else` branch restoring `lastDiceRoll` for RESOLVE_EFFECTS / BUY_OR_BUILD (where the server does hold a non-null roll) is also correct. Field names (`status`, `turnPhase`, `lastDiceRoll`, `roundNumber`) all map cleanly to the server's `GameModel`.

### Tests
The two added tests cover the requested scenarios (stale dice cleared on ROLL_DICE reconnect; `lastDiceRoll` restored on RESOLVE_EFFECTS reconnect). CI + SonarCloud are green.

### Minor (non-blocking) note
The code comment says `lastDiceRoll` is "a stale total from the previous turn" restored by the old code. On the server `lastDiceRoll` is actually `null` in ROLL_DICE, so the old code's `optIntOrNull("lastDiceRoll")?.let { ... }` was skipped — the real source of the stuck Roll button was the client's own retained in-memory `diceResult`, which the new `= null` correctly clears. Behavior is right; only the comment's causal wording is slightly imprecise. Not worth blocking on.

LGTM 🚀

PR #400, opened 2026-06-29
329 parse incoming chat messages

SoupCube:
### Summary
Added an ID check to the chat Message parser `handleChatMessage`
Fixed the `username` fallback when `sender` is missing
Added and adjusted chat tests to include and account for the changes


eles17 (changes requested):
Thanks for the work! I’m requesting two changes before merging:

1. `payload?.optString(...)` returns `""` when a field is missing, so the fallback to top-level `gameId`, `content`, or `sender` is skipped whenever `payload` exists. This causes valid mixed-shape chat messages to be ignored. Please use nonblank fallbacks for these fields.

2. `nonChatMessageDoesNotEmitChatMessage` currently sends `"type":"CHAT"` and duplicates the different-game test. Please change it to use a genuinely non-chat type.

eles17 (approved):
Thanks for fixing the issues. LGTM. Approved

PR #402, opened 2026-06-29
fix: added auto phase change

doniliks:
after entering buying phase delay starts and checks if purchased is pending or finished, if not skips automaticly


Moritz2k1 (approved):
LGTM

PR #403, opened 2026-06-29
fix: clamp dice roll to one die without a Train Station (#399)

valentinaschiavon99:
## Summary

When the active player has **no built Train Station**, rolling a single die wrongly:
- animated **two** dice during the roll, and
- triggered a server error: `NO_TRAIN_STATION` — *"You need a Train Station to roll two dice!"* (`ROLL_FAILED`).

## Root cause

The 1/2 dice-count selector in `DiceSection` is only shown when the player `hasTrainStation`. For a player without one, `selectedDiceCount`/`frozenDiceCount` stay `null`, so both the roll request and the rolling animation fall back to `state.requestedDiceCount`.

`requestedDiceCount` is set on every `rollDice(diceCount)` call and is **never reset back to 1** at the start of a turn (it persists in the ViewModel/state). A stale value of `2` (e.g. carried over from an earlier two-dice roll) therefore leaked into a single-die turn, causing:
- `chosen = 2` → `onRollDice(2)` → server `NO_TRAIN_STATION`, and
- `animationDiceCount = 2` → two dice spin in the UI.

The server (`DiceService`) is correct by the rules — it only requires a Train Station when the resolved dice count is 2 — so the fix belongs on the client.

## Fix

Clamp the effective dice count to **1** whenever the active player has no built Train Station:

- **`GameScreenViewModel.rollDice`** (network boundary): compute `effectiveDiceCount = if (hasTrainStation) diceCount else 1` and use it for both `requestedDiceCount` and `webSocketClient.rollDice(...)`. The server can never receive an illegal two-dice request.
- **`DiceSection`** (animation/selection): the active player's `animationDiceCount` and the roll button's `chosen` value resolve to `1` when there is no Train Station, so only a single die ever spins.

Defense in depth across both layers; a player **with** a Train Station is unaffected and can still choose 1 or 2.

## Tests

Added `GameScreenViewModelTest` coverage:
- `rollDiceClampsToOneDieWhenActivePlayerHasNoTrainStation` — a two-dice request without a Train Station is clamped to 1 (and `requestedDiceCount` becomes 1).
- `rollDiceForwardsTwoDiceWhenActivePlayerHasTrainStation` — with a built Train Station, two dice are still forwarded.

### DIFF
--- a/app/src/main/java/com/machikoro/client/ui/game/GameScreenViewModel.kt
+++ b/app/src/main/java/com/machikoro/client/ui/game/GameScreenViewModel.kt
@@ -338,9 +338,13 @@ class GameScreenViewModel(
         // the dice animation stuck in the rolling state.
         if (mutableState.value.isRolling) return
 
-        mutableState.update { it.copy(isRolling = true, requestedDiceCount = diceCount) }
+        // A player without a built Train Station may only roll a single die. Clamp
+        // here so a stale requested count (e.g. carried over from an earlier two-dice
+        // roll) can never reach the server and trigger NO_TRAIN_STATION (#399).
+        val effectiveDiceCount = if (mutableState.value.hasTrainStation) diceCount else 1
+        mutableState.update { it.copy(isRolling = true, requestedDiceCount = effectiveDiceCount) }
         startRollTimeout(expectedPhase = GamePhase.ROLL_DICE)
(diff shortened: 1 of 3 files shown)


eles17 (approved):
LGTM. Approved.

PR #407, opened 2026-06-30
fix: keep dice rolling UI pending until server response

eles17:
**PR Description**

Fixes the dice roll UI getting stuck in a retry-like flow where the “Roll Dice” button could appear again while the first roll request was still pending.

**What Was Added**

- The dice section now treats `state.isRolling` as part of the visible rolling state.
- While a roll or reroll request is pending, the UI keeps showing the dice rolling animation.
- The normal roll and reroll buttons are disabled/guarded while `state.isRolling` is true.

**Expected Behavior Now**

When the active player taps “Roll Dice”:

- The dice animation starts immediately.
- The UI no longer returns to dice selection just because the short local animation finished.
- The player cannot accidentally trigger additional roll requests while the first one is still pending.
- The animation stays visible until:
  - the server sends a dice result,
  - the server changes the game phase,
  - or the existing 10-second timeout clears the pending roll state.

For Radio Tower rerolls, “Roll Dice Again” follows the same pending-state behavior and cannot be spammed while waiting for the server response.

### DIFF
--- a/app/src/main/java/com/machikoro/client/ui/game/ui/Dice.kt
+++ b/app/src/main/java/com/machikoro/client/ui/game/ui/Dice.kt
@@ -259,6 +259,7 @@ fun DiceSection(
     } else {
         localAnimationDiceCount
     }
+    val showRollingAnimation = isAnimating || (state.isActivePlayer && state.isRolling)
 
     LaunchedEffect(state.diceResult) {
         if (state.diceResult == null) {
@@ -298,7 +299,7 @@ fun DiceSection(
         modifier = modifier.padding(bottom = 32.dp)
     ) {
         when {
-            isAnimating -> {
+            showRollingAnimation -> {
                 Row(horizontalArrangement = Arrangement.spacedBy(16.dp)) {
                     repeat(animationDiceCount) {
                         DiceAnimationDisplay(animating = true)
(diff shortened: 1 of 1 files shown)


valentinaschiavon99 (approved):
LGTM ✅

Traced the full `isRolling` lifecycle — set in `rollDice`/`rerollDice`, cleared on the server `diceResult`, the `diceRollTick` bump, a mismatched phase advance, and the 10s roll timeout — so gating the animation and controls on `showRollingAnimation = isAnimating || (isActivePlayer && isRolling)` correctly holds the rolling UI until the server responds without any risk of getting permanently stuck. The fix is well-scoped and the surrounding comments are clear.

**Non-blocking nit:** the per-button `isRolling` guards are effectively dead code. The roll/reroll buttons only render inside `if (isActivePlayer && gameStatus == IN_PROGRESS && !showRollingAnimation)`, and for the active player `showRollingAnimation` is already true whenever `isRolling` is true. So that block only renders while `isRolling == false`, which means `enabled = !state.isRolling` is always `true` and `if (state.isRolling) return@ActionButton` can never fire. The real concurrent-tap protection is the `if (mutableState.value.isRolling) return` in the ViewModel's `rollDice`/`rerollDice`. Worth either dropping the button-level guards as redundant or leaving a comment that they're intentional belt-and-suspenders — not a blocker either way.

PR #409, opened 2026-06-30
chore: remove unreachable dice rolling guards

eles17:
## Summary
- Remove unreachable `isRolling` checks from the Roll and Reroll buttons in `DiceSection`
- Rely on the existing outer `!showRollingAnimation` gate and ViewModel `isRolling` guard for roll protection

## Testing
- `./gradlew compileDebugKotlin`
- `./gradlew testDebugUnitTest --tests com.machikoro.client.ui.game.GameScreenViewModelTest`

Closes #408

### DIFF
--- a/app/src/main/java/com/machikoro/client/ui/game/ui/Dice.kt
+++ b/app/src/main/java/com/machikoro/client/ui/game/ui/Dice.kt
@@ -385,13 +385,11 @@ fun DiceSection(
                         if (state.hasTrainStation) frozenDiceCount ?: selectedDiceCount ?: state.requestedDiceCount else 1
                     ActionButton(
                         onClick = {
-                            if (state.isRolling) return@ActionButton
                             SoundManager.play(GameSound.DICE_ROLL)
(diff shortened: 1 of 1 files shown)


valentinaschiavon99 (approved):
Verified the core claim: the removed `isRolling` guards were genuinely unreachable.

Both buttons render only inside the `!showRollingAnimation` block, where `showRollingAnimation = isAnimating || (state.isActivePlayer && state.isRolling)`. Since `isActivePlayer` is already true in that branch, `!showRollingAnimation` forces `state.isRolling == false`, so:

- `if (state.isRolling) return@ActionButton` could never trigger — the `onClick` closure reads the same `state` snapshot the gate was computed from.
- `enabled = !state.isRolling` was always `true`, and `ActionButton`'s `enabled` parameter defaults to `true`, so dropping the argument preserves identical behavior.

The remaining gates (outer `!showRollingAnimation` plus the ViewModel-side `isRolling` guard) still protect against double-rolls. Clean removal — LGTM. 👍

PR #414, opened 2026-06-30
refactor: change sizing making space for resolving effects columns

doniliks:
Before: 
Screenshot_20260701_094311

After:
Screenshot 2026-07-01 at 00 14 17


eles17 (changes requested):
The shop’s four-card rows exceed the available center width at the 915dp reference size. The center provides ~607dp, while four 155dp cards plus spacing require 644dp, causing uneven card sizing.
Please reduce the card width or use a four-column grid.

eles17 (approved):
I believe in you Danylo

PR #415, opened 2026-06-30
401 automaticly change buying phase after timeout

doniliks:
Refactored version uses constant delay value and does not call action after phase end


eles17 (changes requested):
The buying timeout and visible countdown can become desynchronized after a failed purchase. PENDING cancels the timeout, and ERROR starts a new 15-second delay, while the displayed countdown continues from its original value.
Please preserve a single phase deadline or reset both timers together. Otherwise, the UI may show 0 while the phase remains active.

eles17 (changes requested):
The single deadline now keeps the timers synchronized, but LaunchedEffect retains the GameScreenState from when it started. If a purchase becomes PENDING during the delay, the final check can still see the old state and advance the turn.

Please use rememberUpdatedState to read the latest state after the delay while retaining the single deadline. Also remove the unused phaseTimerKey and imports.

PR #416, opened 2026-07-01
411 feat animate newly purchased landmarks before ending the purchase phase

Ivona-Goranova:
Adds a landmark purchase reveal before the purchase phase ends.

- Shows all four player landmarks after a landmark purchase
- Animates the newly purchased landmark
- Supports active and passive player views
- Shows a winner message when all landmarks are built
- Adds GameScreen previews and tests
- Improves card sizing, spacing, and timer visibility


eles17 (dismissed):
The landmark reveal can be cut short by the existing 5-second purchase-phase timer, and four 155dp landmark cards exceed the available center-pane width at 915dp, causing clipping or shrinking. Please coordinate the phase transition with the reveal animation and make the landmark row responsive.

Moritz2k1 (approved):
Can be merged according to tohers

PR #417, opened 2026-07-01
fix: Gamescreen

Moritz2k1:
1. GameScreen.kt (line ~541)
   File: app/src/main/java/com/machikoro/client/ui/game/GameScreen.kt

   Problem: Three bugs in the LaunchedEffect caused purchases to be cancelled:
     a) Timer used hardcoded 5000ms but the visible countdown showed 15 seconds.
        Players who bought after second 5 raced against an already-sent endTurn.
     b) Guard condition used || instead of &&, making it always true — the check
        for SUCCESS/PENDING never actually suppressed the auto-endTurn.
     c) state was captured at LaunchedEffect launch (always IDLE), so the live
        purchase state after the delay was never read.

   Fix:
     - delay(5000L) changed to delay(SHOP_VIEW_DELAY) to match the visible timer
     - || changed to && in the purchase state guard
     - Replaced stale var with rememberUpdatedState(state.purchaseState) so the
       coroutine reads live state after the delay, not the launch-time snapshot
     - Added import for rememberUpdatedState

2. GameScreenViewModel.kt (line 45)
   File: app/src/main/java/com/machikoro/client/ui/game/GameScreenViewModel.kt

   Problem: PURCHASE_DISPLAY_DELAY was 5000ms — a mandatory 5-second wait after
   every card buy before the game advanced. This was the biggest cause of the
   "every action takes multiple seconds" feeling. It also caused a double-endTurn:
   both the LaunchedEffect (at 5s) and the ViewModel (at purchase_time + 5s) were
   both calling endTurn, resulting in spurious error toasts.

   Fix:
     - PURCHASE_DISPLAY_DELAY reduced from 5_000L to 1_500L
     - 1.5s is enough to display the purchase result before advancing the turn

### DIFF
--- a/app/src/main/java/com/machikoro/client/ui/game/GameScreen.kt
+++ b/app/src/main/java/com/machikoro/client/ui/game/GameScreen.kt
@@ -580,7 +580,7 @@ fun GameScreen(
                                     verticalAlignment = Alignment.CenterVertically,
                                     modifier = Modifier
                                         .align(Alignment.BottomCenter)
-                                        .offset(y = (-80).dp)
+                                        .offset(y = (-20).dp)
                                 ) {
                                     ActionButton(
                                         onClick = { onReroll(state.diceResult?.size ?: 1) },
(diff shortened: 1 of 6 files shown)


doniliks (commented):
for animations at least 5 sec, otherwise thank you for fixing

